PR #2, opened 2017-11-16
Text color change

Shefali-A:
halp

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -2,4 +2,5 @@ body {
   margin: 0;
   padding: 0;
   font-family: sans-serif;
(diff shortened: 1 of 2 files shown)


src/index.css
```diff
@@ -2,4 +2,5 @@ body {
   margin: 0;
   padding: 0;
   font-family: sans-serif;
+  color: pink;
```
karunsingh: This is what a comment might look like

karunsingh:
Going to close this for now without merging -- but in general, after approval, we would go ahead and merge these pull requests into the master branch

PR #4, opened 2017-11-18
Add basic header content

vanessaw1:


src/SessionInformationHeader.tsx
```diff
@@ -5,7 +5,32 @@ class SessionInformationHeader extends React.Component {
...
+                                <p>Wednesday, 8 Nov</p>
+                                <p>10:00 - 11:00 AM</p>
+                            </div>
+                            <div className="OfficeHourLocation">
+                                G23 Gates Hall
+                            </div>
+                        </div>
+                    </div>
+                </div>
+                <div className="Spacing" />
```
karunsingh: Another way to do this is by setting a margin-top on the button. Try removing the Spacing div entirely, and instead just set margin-top to 10px for .button, that should do it!

src/SessionInformationHeader.css
```diff
@@ -6,7 +6,16 @@
 
 .header { /* Above button */
     background-color: #9b9b9b; 
-    padding: 30px;
+    padding: 20px;
+}
+
+.Close {
```
karunsingh: Hmm, this doesn't seem to work for me -- for now, could we just have a button that has the letter 'x' on it? Don't worry too much about images, we will be doing those once we start implementing high fidelity designs

src/SessionInformationHeader.tsx
```diff
@@ -5,7 +5,33 @@ class SessionInformationHeader extends React.Component {
     render() {
         return(
             <div className="SessionInformationHeader">
-                Session information placeholder
+                <div className="header">
+                    <button className="Close" type="submit">
+                        X
+                    </button>
+                    <div className="CourseInfo">
+                        <span className="CourseNum">CS 3110  </span>
```
karunsingh: Could we now make these parameters into React properties? I just pushed a similar change into my own PR, take a look at how that's to be done:
https://github.com/cornell-dti/office-hours/pull/6/commits/a6af4591b49ae1910adfa2809a8448e59217719a

In your case, we basically want every hard-coded parameter to become a React property (like in my case it was the question text and the number of people) -- which means the class name, professor name, queue size, location, etc should become properties for you. Then, you'll want to change `StudentSessionView.tsx` slightly so that these properties are passed in when creating `<SessionInformationHeader />`. These things will still be hard-coded for now -- just that they'd be hard-coded in the parent component rather than in this file. Let me know if you have questions!

PR #5, opened 2017-11-18
session questions design

Shefali-A:
Frone-end stuff for the session questions section.


src/SessionQuestionsComponent.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import './SessionQuestionsComponent.css';
+
+const peopleLogoImage = require("./peopleLogo.jpg");
```
karunsingh: Change " to '

src/SessionQuestionsComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+                <p className="Name">Karun Singh</p>
+                <p className="Question">How do implement recursion on question 4?</p>
+                <div className="Tags">
+                    <p>Assignment 1</p>
+                    <p>Q4</p>
+                    <p>Recursion</p>
+                    <p>Conceptual</p>
+                </div>
+                <div className="BottomBar">
+                    <img src={peopleLogoImage} className={"peopleLogo"} alt="3 people logo"/>
```
karunsingh: Change " to '

src/SessionQuestionsComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+                    <p className="NumberOfPeople">2</p>
+                    <button className="Button">Resolve</button>
+                </div>
+            </div>
+        )
```
karunsingh: Add semicolon after the closed paren

src/SessionQuestionsComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+import './SessionQuestionsComponent.css';
+
+const peopleLogoImage = require("./peopleLogo.jpg");
+
+class SessionQuestionsComponent extends React.Component {
+    // constructor(props: any) {
+    //     super(props);
+    //     this.state = {name: this.props.name };
+    // }
+
```
karunsingh: Seems like what we were missing was this:
props: {
    name: string
}
(Source: https://www.typescriptlang.org/docs/handbook/jsx.html)
The child just needs to declare what its properties are, and what type they're are. Then we should be able to freely use this.props.name, etc. Try converting the name, question, and number of people to properties. We should then also be able to convert the list of tags to a property (https://reactjs.org/docs/lists-and-keys.html)

karunsingh: Just did it for my own branch, take a look for an example! https://github.com/cornell-dti/office-hours/commit/a6af4591b49ae1910adfa2809a8448e59217719a

karunsingh:
@Shefali-A Once https://github.com/cornell-dti/office-hours/pull/15 is done, let's create a new PR to address the comments above -- it's just a bunch of minor lint things + changing hard-coded values to props, shouldn't take too long!

PR #10, opened 2017-11-21
Added correct schema

Chillee:
Added a proper schema for our data, as well as resolvers for querying.
Check server/db.sql for the SQL schema. It also has some mock data for testing.
Also either check localhost:3001/graphiql for documentation, or check out `server/schema.gql`.

Here's an example of the API in action: .
Here's a diagram of the db schema 

If you have any ideas for useful functionality in the API, please speak up.


server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    name varchar,
+    semester varchar
+);
+
+create table students (
+    netid varchar PRIMARY KEY,
+    name varchar
+);
+
+CREATE TABLE course_students (
```
karunsingh: This probably isn't needed, since everyone is enrolled in everything by default -- for now. Maybe to reduce technical debt we should still just have it in the database and populate it redundantly?

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    course_id integer,
+    netid varchar,
```
karunsingh: Change to `student` to be consistent with the `course_tas` field naming below. Alternatively change both to `student_netid` and `ta_netid`

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    PRIMARY KEY(course_id, netid),
+    FOREIGN KEY(course_id) REFERENCES courses(course_id),
+    FOREIGN KEY(netid) REFERENCES students(netid)
+);
+
+CREATE TABLE course_tas (
+    course_id integer,
+    ta varchar,
```
karunsingh: See above

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    PRIMARY KEY(course_id, ta),
+    FOREIGN KEY(course_id) REFERENCES courses(course_id),
+    FOREIGN KEY(ta) REFERENCES students(netid)
+);
+
+CREATE TABLE sessions (
+    session_id integer PRIMARY KEY AUTOINCREMENT,
+    time varchar,
```
karunsingh: First, let's change this to an integer if we're storing unix time

karunsingh: Second, it's a little bit more involved because each session will have a start time and an end time. There's two alternatives; one is to store the start time, and then a `length` which implicitly gives you the end time. Or store the `end_time` as its own field instead, which implicitly gives you the length. I don't see any issues with either, so that's entirely up to you.

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    location varchar,
+    course_id integer,
+    FOREIGN KEY(course_id) REFERENCES courses(course_id)
+);
+
+CREATE TABLE session_tas (
+    session_id integer,
+    ta varchar,
```
karunsingh: Same as one of the comments above -- let's just choose a consistent convention for fields that reference NetIDs as a foreign key

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    PRIMARY KEY(session_id, ta),
```
karunsingh: Nice, this should allow for multiple TAs per session then 👍

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    FOREIGN KEY(session_id) REFERENCES sessions(session_id),
+    FOREIGN KEY(ta) REFERENCES students(netid)
```
karunsingh: This should reference `course_tas` rather than `students`

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+);
+
+CREATE TABLE questions (
+    question_id integer PRIMARY KEY AUTOINCREMENT,
+    text text,
```
karunsingh: Could we change the name of the field to `content` or something? Using the reserved keyword can sometimes lead to weird errors

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    time_entered datetime DEFAULT CURRENT_TIMESTAMP,
```
karunsingh: Hm I didn't know datetime was supported by SQLite? Either way, can we change this to Unix-style (integer) for consistency?

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    session_id integer,
+    student varchar,
+    FOREIGN KEY(session_id) REFERENCES sessions(session_id),
+    FOREIGN KEY(student) REFERENCES students(netid)
+);
+
+CREATE TABLE question_followers (
+    question_id integer,
+    follower integer,
```
karunsingh: Should be a string

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    PRIMARY KEY (question_id, follower),
+    FOREIGN KEY(question_id) REFERENCES questions(question_id),
+    FOREIGN KEY(follower) REFERENCES students(netid)
+);
+
+CREATE TABLE tags (
+    tag_id integer PRIMARY KEY AUTOINCREMENT,
+    value text,
```
karunsingh: Change to something like `tag_value` to be more specific

server/db.sql
```diff
@@ -0,0 +1,136 @@
+PRAGMA foreign_keys = ON;
+
+CREATE TABLE courses (
+    course_id integer PRIMARY KEY AUTOINCREMENT,
+    name varchar,
```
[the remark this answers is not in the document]
Chillee: I don't think there's many benefits on the db side. We won't need to search by department or w.e.

If you're worried about front end ease, I can add some entries to GraphQL,

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+    semester varchar
```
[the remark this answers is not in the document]
Chillee: I think that makes the most sense.

server/db.sql
```diff
@@ -0,0 +1,136 @@
...
+);
+
+create table students (
```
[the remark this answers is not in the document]
Chillee: good catch

server/db.sql
Chillee: yep, that's why I kept it here. It's something easy to add that we'll need in the future.

It shouldn't really affect anything else.

server/db.sql
Chillee: hmmm. Alright I think that makes sense.

server/db.sql
Chillee: yep, this was a holdover from previously that I forgot to change. I'll probably do start/end time.

server/db.sql
Chillee: hmm... I seem to remember having a good reason for choosing `students` instead of `course_tas`...

Ah yes. I think the reasoning was that a TA for a session might not necessarily be an "official ta" for any course. I also thought that it didn't really make sense to have this kind of foreign key constraint.

Oh yes. The real reason is that we are unable to do it. Course_tas is a joint table, and tas is not unique. You can only do foreign keys to a primary key (or at least an unique key).

server/db.sql
Chillee: sure

server/db.sql
Chillee: DATETIME is basically just a cast into an integer, I believe.
image

server/db.sql
Chillee: nice catch

server/db.sql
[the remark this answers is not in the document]
karunsingh: I'm thinking from a flexibility point of view -- splitting them up gives us everything that combining does, and more

server/db.sql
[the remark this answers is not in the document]
karunsingh: Sounds good, could we just have it documented as a comment here if possible?

server/db.sql
karunsingh: Alright, just need to make sure that we populate it

server/db.sql
karunsingh: Hmm this might be too strong a constraint. Would professors be entered as users in the students table?

server/db.sql
karunsingh: Okay, then let's make all of them `DATETIME`

karunsingh:
We also need to figure out how we're representing professors. Also, we need to distinguish between TAs and 'privileged' course staff who can make admin changes (for example, professors and head TAs)

server/db.sql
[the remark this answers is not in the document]
Chillee: Not really, imo. It doesn't give us anything we want, and it just incurs more complexity. If we ever need to split it up, it shouldn't be too hard to do so.

server/db.sql
Chillee: I'm thinking that we would just add another table called session_admins and session_profs.

server/db.sql
Chillee: I don't think it's necessary since it's already in the tags table.

Chillee:
@karunsingh See my above comments about the `course_tas` table.

server/db.sql
karunsingh: Why session? I think we should add a field to `course_tas` indicating an `admin` role (true or false) because `admin` roles are specific to courses rather than sessions.

karunsingh: That, coupled with changed `students` to simply `users` (and entering all of our users there) should do the trick

Chillee: *yeah I meant `course_admins` not `session_admins`

server/db.sql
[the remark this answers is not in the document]
karunsingh: Okay. If and when this expands to multiple courses, we'll definitely want some department filters -- but you're right, shouldn't be hard to split later. I do think it's worth adding a field for the course title though. I.e. "CS 3110", "Functional Programming" should be separate fields. We'll need to use them separately in some places I believe

karunsingh:
@Chillee See responses; I also just remembered that we need to add some fields to the `questions` table regarding question resolution. Most importantly, a flag indicating that the question has been resolved, and secondary is the time at which it was resolved

server/db.sql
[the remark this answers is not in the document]
Chillee: I think stuff like providing them separately is better handled in the GraphQL layer.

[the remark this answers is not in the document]
karunsingh: Will you be splitting strings to separate them?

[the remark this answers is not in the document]
Chillee: in the actual GraphQL layer, yes.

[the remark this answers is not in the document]
karunsingh: Hm, I don't understand the advantage of doing that - what am I missing? Especially because "CS 3110 Functional Programming" will never be needed as one combined field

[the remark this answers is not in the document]
Chillee: Ok. I think for this case, the "Functional Programming" part is not an inherent part of identifying the course, and can be split.

PR #14, opened 2017-11-27
Add first version of Calendar Header

vanessaw1:


client/src/styles/CalendarHeader.css
```diff
@@ -1,3 +1,22 @@
 .CalendarHeader {
-    border: 1px solid #000000;
+    font-size: 20px;
+    font-weight: bold;
+    padding: 0px 15px;
```
karunsingh: Let's add some padding-top, something like 10px

client/src/components/includes/CalendarHeader.tsx
```diff
@@ -2,10 +2,22 @@ import * as React from 'react';
...
+        currentCourse: string
+    };
+
     render() {
         return (
             <div className="CalendarHeader">
-                Calendar header placeholder
+                <div className="CurrentCourse">
+                    {this.props.currentCourse}
+                    <button className="CourseSelectButton">
```
karunsingh: This works, but let's try and use the [Semantic-UI icon button](https://react.semantic-ui.com/elements/button#button-example-icon-prop) instead of `<button><i /></button>` -- it'll make some React stuff easier later on. If you use that, then we should be able to use the [size property](https://react.semantic-ui.com/elements/button#button-example-size) to increase/decrease button sizes

karunsingh: If that causes some weird issues, we can go ahead with this version for now! To increase or decrease icon sizes, you can treat them as normal fonts. Changing their `font-size` should work!

vanessaw1:
Changed the buttons. The size property seems to change the padding much more than the icon size itself, so I couldn't get them as large as I wanted. I'm also having trouble with getting custom colors to show up.

karunsingh:
@vanessaw1 You're right, let's revert back to the version without Semantic-UI. We can keep the padding change, and to increase/decrease the size of the icons, you should be able to change their font-size

client/src/components/includes/CalendarHeader.tsx
```diff
@@ -12,9 +11,13 @@ class CalendarHeader extends React.Component {
             <div className="CalendarHeader">
                 <div className="CurrentCourse">
                     {this.props.currentCourse}
-                    <Button className="SelectButton" size="big" compact icon="angle down" />
+                    <button className="CourseSelectButton">
+                        <i className="angle down icon" />
+                    </button>
                 </div>
-                <Button className="MenuButton" size="big" compact icon="bars" />
+                <button className="MenuButton">
```
karunsingh: Let's make this one's font-size 25px as well to have it line up with the stuff on the left -- we can refine it later. After changing that, we can merge this!

PR #15, opened 2017-12-02
changes to add question

Shefali-A:
Rough layout of the design.


client/src/components/includes/QuestionHeader.tsx
```diff
@@ -0,0 +1,93 @@
...
+        var primaryTagsList = this.props.primaryTags.map(
+          (tag, index) => {
+            return <SelectedTags index={index} tag={tag} ifSelected={this.state.primaryBooleanList[index]} onClick={this.handleSelected}/>
+          }
+        );
+
+        var secondaryTagsList = this.props.secondaryTags.map(
+          (tag, index) => {
+            if (this.state.secondaryBooleanList[index]) {
+              return <p className="selectedTag" key={index} onClick={this.handleSelected}>{tag}</p>
```
karunsingh: Looks like the error you're getting is because of the secondary, and not the primary tags! If you change the `<p>` element here to a `<SelectedTag>` element, it should work. You'll probably need to write another `handleSelected` function though for the secondary tags, since the current one will only toggle the primary ones.

karunsingh:
On Sunday, could you start by `git pull`ing `master` into your branch? There will be a tiny conflict in `App.tsx`, but we can resolve it pretty easily!

client/src/styles/QuestionHeader.css
```diff
@@ -0,0 +1,86 @@
+.QuestionHeader {
```
karunsingh: This file and `QuestionView.css` can be removed now!

client/src/components/App.tsx
```diff
@@ -10,14 +11,15 @@ class App extends React.Component {
       <Router>
         <div className="App">
           Navigate: | <Link to="/calendar">Calendar View</Link> | <Link to="/session">Student Session View</Link> |
-          <Link to="/session-ta">TA Session View</Link> |
+          <Link to="/question">Question View</Link> | <Link to="/session-ta">TA Session View</Link> |
           <Route path="/calendar" component={CalendarView} />
           <Route path="/session" component={StudentSessionView} />
+          <Route path="/question" component={QuestionView} />
```
karunsingh: Now that I think about the naming, it's probably better to name the component `AddQuestionView` instead of `QuestionView` and have the route be `/add-question`, just so that no one is confused

client/src/components/includes/QuestionHeader.tsx
```diff
@@ -0,0 +1,149 @@
+import * as React from 'react';
+import SelectedTags from '../includes/SelectedTags';
+
+class QuestionHeader extends React.Component {
```
karunsingh: Could we change all instances of `QuestionHeader` to `AddQuestion` wherever it occurs? (same reasoning as before, it'll make the code easier to understand)

PR #18, opened 2017-12-02
Webpack, Less, file moving

SHBelsky:


.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
```
karunsingh: Could we keep these VS Code settings?

README.md
```diff
@@ -1,14 +1,18 @@
 # Office Hours
-./client contains the development server for the react app, while ./server contains the GraphQL server for the backend
+`./client` contains the development server for the react app, 
+while `./server` contains the GraphQL server for the backend
 
 ## Setup
-Run
+
+1. Install Yarn [here](https://yarnpkg.com/en/docs/install).
```
karunsingh: Why do we need yarn?

karunsingh: (Can we stick to npm?)

karunsingh:
@SHBelsky As far as adding a CSS pre-processor, it looks like create-react-app has some instructions of its own: https://github.com/cornell-dti/office-hours/tree/master/client#adding-a-css-preprocessor-sass-less-etc

SHBelsky:
I'm in Israel at the moment, so I'm away from my computer until Saturday. I'll revisit this when I'm back.

PR #20, opened 2017-12-17
Add first version of Calendar Sessions

vanessaw1:
Some issues:
* It looks like I'll need to include some Javascript to convert unix times to regular times, but I'm not sure how to include that code in the .tsx file
* I couldn't get the SessionIndicator (the red/white circles) to highlight only for the first session. Though eventually they'll be highlighted based on if the OH is occuring right now.
* It looks like the line lengths are based on how long the OH is. I'm not sure how to dynamically change the CSS based on that.
* The dots and lines are centered manuallly since I couldn't find a way to center elements in different dividers.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -3835,6 +3835,19 @@
         "minimist": "0.0.8"
       }
     },
+    "moment": {
+      "version": "2.20.0",
+      "resolved": "https://registry.npmjs.org/moment/-/moment-2.20.0.tgz",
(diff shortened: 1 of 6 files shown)


karunsingh:
This looks great so far! We don't need to worry about the line lengths and the manual centering, the fine details can come in later. I'm just going to push a change to this that adds a JS library for handling these timestamps, which can then handle the first two issue

karunsingh:
@vanessaw1 Check out the [changes](https://github.com/cornell-dti/office-hours/pull/20/commits/11687344ea98ec459af55ea3dc06b7f80f8ed1dd) I just added, and let me know if you have questions! One thing is that we will probably add different indicators for office hours that are ongoing (live), and those that are open for queuing -- there should be an updated design soon, but I've made the distinction in the code for now.

We can merge this whenever you're ready!

PR #22, opened 2017-12-21
Setting up Apollo

karunsingh:
- Install Apollo and integrate it with the React tree
- Change CalendarSessions to pull data from the database via GraphQL


karunsingh:
@Chillee Take a look, there's some minor changes to the backend:
- Whitelisting the frontend to bypass CORS
- SQL linting
- Changing some of the initialization data
- Changing the Sessions gql to have `start` and `end` fields instead of `time`

PR #23, opened 2017-12-23
Setting up react-router

karunsingh:
- Installed `react-router-dom`
- Separated Calendar and Student Session views to their own routes
- Added navigation at the top for convenience

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -4,6 +4,11 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


karunsingh:
@vanessaw1 @Shefali-A Take a look at the file changes in App.tsx: `react-router` just helps us separate out the different views onto their own screens using URL routes. Now, for example, the calendar view will be at localhost:3000/calendar, and the student session view will be at localhost:3000/session. Let me know if you have questions!

PR #26, opened 2018-01-06
Detailed Question View (initial)

vanessaw1:
- The background stops scrolling when the pop-up is visible, but the scroll position jumps up to the top.
- I'm passing down the toggle function to each subcomponent using prop; let me know if there's a better way to do this.
- I'm also running into a minor issue where some CSS styles are unintentionally applied to other HTML files, such as if class names are the same, or if I'm styling general elements like buttons. I can get around this by using unique class names, but is there a way to make a CSS file specific to one .tsx file?
- How should information from a question card be passed to the pop-up?

### DIFF
--- a/client/src/components/includes/DetailedQuestionView.tsx
+++ b/client/src/components/includes/DetailedQuestionView.tsx
@@ -3,11 +3,22 @@ import '../../styles/DetailedQuestionView.css';
 
 class DetailedQuestionView extends React.Component {
     props: {
+        isDetailed: boolean,
         studentName: string,
         studentQuestion: string,
         tags: string[],
-        group: string[]
+        group: string[],
+        handleClick: Function
     };
+
+    constructor(props: {}) {
+        super(props);
+        this._toggleDetails = this._toggleDetails.bind(this);
+    }
+
+    _toggleDetails(prev: boolean) {
+        this.props.handleClick(prev)
+    }
 
     render() {
         var tagsList = this.props.tags.map(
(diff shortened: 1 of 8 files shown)


client/src/components/includes/DetailedQuestionView.tsx
```diff
@@ -0,0 +1,65 @@
...
+        super(props);
+        this.toggleDetails = this.toggleDetails.bind(this);
+    }
+
+    toggleDetails(prev: boolean) {
+        this.props.handleClick(prev);
+    }
+
+    render() {
+        var tagsList = this.props.tags.map(
```
karunsingh: Great use of `map` :)

karunsingh:
Looks good!
- Background scrolling: for now, this is fine -- we can spend time fixing it if need be later
- Passing down the toggle function as a prop is the way to do it!
- Yeah this is a recurring problem, and we need to fix it soon. For now, the only solution is having unique class names for each component. One simple way is to just prepend the class names with the component name, like "component-name-class-name" instead of just "class-name". We definitely need a more permanent fix
- We can talk more about this in-person next week, but the gist of it is that we can store not only `this.state.isDetailed` in `StudentSessionView`, but also _which_ question was clicked for the detailed view popup in a second state variable (for eg, `this.state.detailedQuestionID`). Then, inside of `render()` for `StudentSessionView`, we can use `this.state.detailedQuestionID` to fetch the relevant question's details and pass them as props to `DetailedQuestionView`. Let's discuss this in person, and once the change is in, we can wrap up this PR!

PR #28, opened 2018-01-27
Linking calendar sessions to session view

vanessaw1:


client/src/components/includes/CalendarSessionCard.tsx
```diff
@@ -51,14 +74,14 @@ class CalendarSessions extends React.Component {
                 </div>
                 <div className="CalendarInfo">
                     <div className="SessionDivider" />
-                    <div className="CalendarCard">
+                    <div className="CalendarCard" onClick={this.handleOnClick}>
```
karunsingh: One minor edit: could we change the cursor to the `pointer` cursor for the cards? (just add `cursor: pointer` to the CSS for this class)

PR #30, opened 2018-02-05
Vw/pass info from card

vanessaw1:
Was able to move the DetailedQuestion component down to SessionQuestionContainer, but popular questions are not clickable anymore because it can't see and pass info down to the DetailedQuestion component. We could have a parent div enclosing SessionQuestionsContainer and PopularQuestionsContainer so that DetailedQuestion can be made from there.

### DIFF
--- a/client/src/components/includes/DetailedQuestionView.tsx
+++ b/client/src/components/includes/DetailedQuestionView.tsx
@@ -33,6 +33,11 @@ class DetailedQuestionView extends React.Component {
             }
         );
 
+        var groupListEmpty = 'Empty';
+        if (groupList.length > 0) {
+            groupListEmpty = '';
+        }
+
         var popup = 'PopupInvisible';
         if (this.props.isDetailed) {
             popup = 'PopupVisible';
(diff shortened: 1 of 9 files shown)


client/src/components/includes/SessionQuestionsComponent.tsx
```diff
@@ -20,22 +26,26 @@ class SessionQuestionsComponent extends React.Component {
 
     toggleDetails(prev: boolean) {
         this.props.handleClick(prev);
+        this.props.updateDetails(this.props.studentName, this.props.studentQuestion, this.props.tags, this.props.group);
```
vanessaw1: This is where info from a card is passed up to the detailed question component. Currently it's passing info directly, but can probably change this method to only pass an ID later on.

client/src/components/includes/DetailedQuestionView.tsx
```diff
@@ -33,6 +33,11 @@ class DetailedQuestionView extends React.Component {
             }
         );
 
+        var groupListEmpty = 'Empty';
```
karunsingh: This CSS approach works, but looks like we can make it cleaner using just JavaScript: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

See if you can get it working the same way without the CSS hack!

PR #31, opened 2018-02-09
LESS with react-scripts

SHBelsky:
Allows us to develop CSS with LESS preprocessor, without needing to eject or configure webpack on our own.


karunsingh:
@SHBelsky Does the `less-watch-compiler` integration look alright? Let me know if you have any concerns!

SHBelsky:
@karunsingh Do we need to be watching for changes to CSS if we are committing a built version of the LESS? `build-less` won't stop and allow the next part of `npm start` to execute. You would need to execute those commands with `concurrently` so that the backend server starts *and* LESS is compiled and watched (if changes need to be made.)

karunsingh:
Hmm, I think having it watch for changes is very useful in development and
of course not useful in the deployment. I was thinking that:
- if we run the client on its own, we can watch for changes since this will
only happen in development
- in deployment, `npm run build` is called at the top-level so that the
backend can serve a static front-end. In that, we can use just lessc
instead of watching for changes. What do you think?
Alternatively, running the client can result

On Feb 25, 2018 7:23 PM, "Shea Belsky" <notifications@github.com> wrote:

@karunsingh <https://github.com/karunsingh> Do we need to be watching for
changes to CSS if we are committing a built version of the LESS? build-less
won't stop and allow the next part of npm start to execute. You would need
to execute those commands with concurrently so that the backend server
starts *and* LESS is compiled and watched (if changes need to be made.)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/cornell-dti/office-hours/pull/31#issuecomment-368359564>,
or mute the thread

SHBelsky:
Will the front end developers be able to work on the app without the back end functioning (running it headless, in a way?) If that's the case, then I'm OK with merging it with your edits.

karunsingh:
Yup, that is the intention. I'm working on another PR that will let the client mock backend queries in the absence of an actual backend, which will enable that kind of headless development.

PR #34, opened 2018-02-11
Initial temporary login component for Oauth

Glitched:
Quick component to get a Oauth response from Google, will be used to authenticate users and store their status.


client/src/components/pages/LoginView.tsx
```diff
@@ -2,9 +2,9 @@ import * as React from 'react';
 import GoogleLogin from 'react-google-login';
 
 class LoginView extends React.Component {
-  responseGoogle = (response: any) => {
+  responseGoogle = (response: ReactGoogleLogin.GoogleLoginResponse) => {
```
karunsingh: Nice! Did they expose some types in the package?

Glitched: Yeah! I had a hard time finding it earlier because it wasn't in the /src/ directory and stuffed with the other stuff like licenses and config files

PR #36, opened 2018-03-04
Adjust server routes to work in production

karunsingh:
- `serviceWorker.js` lets routes that start with `__` to pass through to the server instead of going to a 'fallback URL' on the client
- Changed all `/auth` endpoints to `/__auth`
- Tested on Heroku


karunsingh:
@Glitched I think I covered all the test cases, but could you also verify the login flow on Heroku?

PR #45, opened 2018-04-06
Sa/student session design

Shefali-A:
Done with both views.


client/src/components/includes/SessionJoinButton.tsx
```diff
@@ -4,10 +4,11 @@ class SessionJoinButton extends React.Component {
     render() {
         return (
             <div className="SessionJoinButton">
-                <span>+</span>
+                <p className="Plus">+</p>
```
karunsingh: Could this be one <p> element with the plus inside a <span>? With two <p>'s, the responsiveness for bigger/smaller screens doesn't hold well

client/src/components/includes/SessionQuestionsComponent.tsx
```diff
@@ -35,19 +36,32 @@ class SessionQuestionsComponent extends React.Component {
...
-                    <button className="Button">Resolve</button>
+                    <p className="Order">{this.props.order}</p>
+                    <p className="Time">{this.props.times}</p>
                 </div>
+                { this.props.isTA ?
+                  <div className="Buttons">
+                    <hr/>
+                    <div className="Button">
+                      <p className="Delete">X Delete</p>
+                      <p className="Resolve">	&#10004; Resolve</p>
```
karunsingh: Let's use `&cross;` for the X and `&check;` for the check

client/src/components/includes/SessionInformationHeader.tsx
```diff
@@ -51,13 +51,12 @@ class SessionInformationHeader extends React.Component {
                 <div className="MoreInformation">
                     <hr/>
                     <div className="QueueInfo">
-                        <img src="../../media/peopleLogo.jpg"/>
+                        <i className="users"></i>
```
karunsingh: Try `<Icon name='users' />` instead, and `import { Icon } from 'semantic-ui-react'` at the top!

PR #48, opened 2018-04-15
Created several PGSQL functions

Chillee:
Namely
```
add_question_with_tags(content text, status text, session_id integer, asker_id integer, tags integer[]) RETURNS SETOF public.questions
create_sessions_from_session_series(series integer) RETURNS SETOF public.sessions
search_session_range(course integer, begintime timestamp without time zone, endtime timestamp without time zone) RETURNS SETOF public.sessions
```
Todos:
- [x] Add an `edit_session_series` function that would take in the updated details of a session series and manipulate all the (subsequent) instantiations of it
- [x] Add a `delete_session_series` function that would delete all subsequent instantiations of the series
- [ ] Add a `edit_session` function that would take in the updated details of a session, and would manipulate it to be independent of the series


package.json
```diff
@@ -1,4 +1,29 @@
 {
+<<<<<<< HEAD
```
karunsingh: Seems to be a merge conflict here, do you mind resolving this one? For package.json, could you also incorporate this small change from https://github.com/cornell-dti/office-hours/pull/46?

server/office_hours.sql
```diff
@@ -48,7 +48,51 @@ SET default_tablespace = '';
 SET default_with_oids = false;
 
 --
--- Name: sessions; Type: TABLE; Schema: public; Owner: -
+-- Name: questions; Type: TABLE; Schema: public; Owner: chilli
```
karunsingh: Could we export this to have no owner? There's a no-owner flag available when dumping the schema -- without it, it throws warnings when it can't find the role 'chilli'

server/office_hours.sql
```diff
@@ -407,6 +485,8 @@ COPY public.questions (question_id, content, time_entered, status, time_resolved
 3	Clarifying statistics concept from prelim	2018-03-26 10:03:12	unresolved	\N	1	7	\N
 1	How do you implement recursion in Question 2?	2018-03-26 09:47:33	resolved	2018-03-26 10:06:49	1	2	1
 4	Question about course grading	2018-03-26 10:07:39	unresolved	\N	1	5	\N
+7	help!!	2018-04-15 14:03:07.622829	unresolved	\N	1	1	\N
```
karunsingh: For now, could we remove these two mock questions?

server/office_hours.sql
```diff
@@ -449,15 +529,27 @@ COPY public.session_tas (session_id, user_id) FROM stdin;
...
-8	2018-04-03 17:00:00	2018-04-03 17:30:00	Upson	B60	\N	1
+COPY public.sessions (session_id, start_time, end_time, location, session_series_id, course_id) FROM stdin;
+1	2018-03-26 10:00:00	2018-03-26 11:00:00	\N	1	1
+143	2018-01-22 12:20:00	2018-01-22 13:10:00	Academic Surge A Office 101	2	1
+3	2018-03-26 13:00:00	2018-03-26 14:30:00	Rhodes 412	3	\N
+144	2018-01-29 12:20:00	2018-01-29 13:10:00	Academic Surge A Office 101	2	1
+4	2018-03-26 19:00:00	2018-03-26 20:30:00	\N	4	\N
+5	2018-04-02 10:00:00	2018-04-02 11:00:00	\N	1	1
+7	2018-04-02 19:00:00	2018-04-02 20:30:00	\N	4	\N
+145	2018-02-05 12:20:00	2018-02-05 13:10:00	Academic Surge A Office 101	2	1
```
karunsingh: It doesn't matter much, but I think it would help if we repopulated the mock sessions from scratch just to ensure the data is consistent with our model

karunsingh:
Added some todos to the top and some small comments to the file changes, but looks great so far!!

server/office_hours.sql
Chillee: I just deleted the session_series_id from the original mock data. They can represent the ad hoc sessions generated.

server/office_hours.sql
```diff
@@ -244,12 +247,21 @@ CREATE TABLE public.session_series (
     session_series_id integer NOT NULL,
     start_time timestamp without time zone NOT NULL,
     end_time timestamp without time zone NOT NULL,
-    building text NOT NULL,
-    room text NOT NULL,
+    location text NOT NULL,
```
karunsingh: We should keep the building and room separate!

Chillee:
@karunsingh Should the delete_session_series also delete the session_series from the table?

karunsingh:
@Chillee Just a couple of changes and we should be good to close this:
- Let's separate out the `building` and `room` fields in `location` (see comment above)
- Discuss the `edit_single_session` function and decide on whether we should leave it to the client or not

(We can have the auth stuff in a new branch / PR, these functions are more time-sensitive)

karunsingh:
@Chillee Let's add a function that returns all the relevant details (`start`, `end`, `building`, `room`, etc) given a session ID. Right now that client is unnecessarily repeatedly handling logic that looks inside the session object as well as the associated session series to get these details -- would be great if we can factor this out into the backend as a single query.

server/office_hours.sql
```diff
@@ -62,14 +102,104 @@ CREATE TABLE public.sessions (
 );
 
 
+--
+-- Name: create_sessions_from_session_series(integer); Type: FUNCTION; Schema: public; Owner: -
+--
+
+CREATE FUNCTION public.create_sessions_from_session_series(series integer) RETURNS SETOF public.sessions
```
karunsingh: Could we change this to also just insert a new session series? It seems like the only use case will immediately follow the insertion of a new session series, so it'd be easier for us to bundle that into the same function call.

PR #49, opened 2018-04-15
Rms/split view

Glitched:
Merge Student Sessions View, TA Sessions View, and Calendar View into unified Split view.

There's some refactoring to be done, but that might be best after BOOM.

Also adds desktop Login UI.

This also has had the connected add question view.

**TODO:** isTa requires courseId to be attached to all sessions


client/src/components/App.tsx
```diff
@@ -1,25 +1,22 @@
...
-                    <Route path="/session/:sessionId" component={StudentSessionView} />
+                    <nav>
+                        <Link to="/login"> Login View</Link> |
+                        <Link to="/question"> Question View</Link> |
+                        <Link to="/desktop/1"> Split View</Link>
+                    </nav>
                     <Route path="/question" component={QuestionView} />
-                    <Route path="/session-ta/:sessionId" component={TASessionView} />
                     <Route path="/login" component={LoginView} />
+                    <Route path="/desktop/:courseId" component={SplitView} />
```
karunsingh: We should probably rename the path from `/desktop` to something more generic, or maybe even have it be the root (`/:courseId`)?

client/src/components/pages/SplitView.tsx
```diff
@@ -96,6 +123,18 @@ class CalendarView extends React.Component {
...
 
+    handleSessionClick = (sessionId: number) => {
+        this.setState((prevState) => {
+            return { sessionId: sessionId, activeView: 'session' };
+        });
+    }
+
+    handleBackClick = () => {
+        this.setState((prevState) => {
+            return { activeView: 'calendar' };
```
karunsingh: Should we also set `sessionId: -1`? The effect will be that hitting the back button on desktop will take them back to the 'select an office hour' screen - right not the back button doesn't give them any feedback

PR #52, opened 2018-04-30
Add Question support for Split View

Glitched:
I added support for the AddQuestion modal flow from Zeplin. The design isn’t in full compliance, but it’s pretty close for now. The mobile fallback does not display as a modal but a full view.


karunsingh:
Looks great to me -- is there any added code here that might need refactoring/revisiting in the future? Besides that, I think we're good to merge!

karunsingh:
Quick question, not fully related to this PR: what does the -1 indicate when the split view URL is `/course/-1`?

PR #59, opened 2018-05-14
Connected Professor View and Tags

vanessaw1:
Allows for viewing and editing office hours

**Notes:**
- User may need to take into account time zone differences, if creating an OH outside New York time
- Popups may be needed for: OH info not filled in all the way, creating/editing an OH with end time already passed, and deleting an OH that has already started
- Automatic refreshing needed in future update
- Professor Tags is currently nonfunctional (based on old design)


client/src/components/includes/ProfessorOHInfo.tsx
```diff
@@ -273,6 +274,24 @@ class ProfessorOHInfo extends React.Component {
...
         );
 
+        // Disable saving if fields are empty
+        // Disable saving if end time (state) is in the past
+        // Disable saving if default end time (prop) is in the past
+        var disableEmpty = this.state.startTime == null && this.state.endTime == null;
+        var disableState = this.state.endTime !== null && moment(this.state.endTime).isBefore();
+        var disableProps = this.props.endTimeDefault !== null && moment(this.props.endTimeDefault).isBefore();
+
+        var tooltip = '';
```
karunsingh: Instead of a tooltip, could we make it a red error message (just like you did in the delete modal)? The user might not see a tooltip, so it's better to play it safe and show it to them explicitly!

vanessaw1: Sure -- should I just put the messages next to the button?

PR #60, opened 2018-06-18
Database Documentation

karunsingh:
- Adding detailed documentation for the database schema and functions in `/docs`
- Redoing PL/pgSQL functions to make them cleaner, consistent, and to separate API and internal functions. Created API functions to manipulate sessions and session series.
- Redid mock data to comply with the documentation/constraints


Glitched:
Looks really good!
Does time_addressed get updated if the question is retracted? Should it?

>(timestamp without time zone): contains the weekly start day and time
Do you think it would be worth while to provide an example value for these? We (you) do it for building/room, but this seems like it has more room for confusion. (I haven't really worked with Postgres, so this might be super obvious

karunsingh:
@Glitched done! Also added in sorting for `api_get_sessions`

PR #61, opened 2018-06-24
Completing Authentication Flow

karunsingh:
- Added backend functions to create/find the logged in user's details
- Completed JWT/Cookie flow
- Implemented PrivateRoutes to protect certain views behind login
- Removed mock/constant userIds in several places, replaced with the relevant backend queries (only the backend should be trusted with retrieving the id of the logged in user)
- Added support for bypassing actual login and assuming user identities for local testing
- Added setup documentation for new developers
- Added documentation about the authentication flow
- Miscellaneous front-end bug fixes

TODO:
- [x] Update database documentation
- [x] Write documentation about the auth flow
- [x] Write documentation about local development setup (with and without login)
- [x] Force getCurrentUser API function to bypass cache


client/src/components/includes/AddQuestion.tsx
```diff
@@ -320,7 +310,7 @@ class AddQuestion extends React.Component {
                             />
                             : <p className="placeHolder text">Finish selecting tags...</p>}
                     </div>
-                    <Mutation mutation={ADD_QUESTION}>
+                    <Mutation mutation={CREATE_QUESTION}>
```
Glitched: Since the api is `apiAddQuestion`, do you think there's value in keeping this consistent (i.e. using `ADD_QUESTION` and `addQuestion` instead)?

client/src/components/includes/CalendarHeader.tsx
```diff
@@ -46,8 +49,9 @@ class CalendarHeader extends React.Component<ChildProps<InputProps, Response>> {
                 <div className="CurrentCourse">
                     {this.props.currentCourse}
                     {data
-                        && data.courseByCourseId
-                        && data.courseByCourseId.courseUsersByCourseId.nodes[0].role === 'ta'
+                        && data.apiGetCurrentUser
+                        && (data.apiGetCurrentUser.nodes[0].courseUsersByUserId.nodes[0].role === 'ta'
+                            || data.apiGetCurrentUser.nodes[0].courseUsersByUserId.nodes[0].role === 'professor')
```
Glitched: Just for consistency, do you think you could indent 54 like the other conditions (so `||` is directly below/above `&&`)?

client/src/components/includes/ConnectedSessionQuestions.tsx
```diff
@@ -102,7 +113,9 @@ class ConnectedSessionQuestions extends React.Component<ChildProps<InputProps, R
                         time: new Date(node.timeEntered),
                         tags: questionTags,
                         userId: node.userByAskerId.userId,
-                        timeEntered: node.timeEntered
+                        timeEntered: node.timeEntered,
+                        photoUrl: node.userByAskerId.photoUrl ||
+                            'https://upload.wikimedia.org/wikipedia/commons/7/7c/Profile_avatar_placeholder_large.png'
```
Glitched: Could we move this into static just in case this image isn't around on wikimedia forever? (Assuming license permits, if not we can kick this over to the designers)

client/src/components/includes/SessionInformationHeader.tsx
```diff
@@ -101,6 +77,7 @@ class SessionInformationHeader extends React.Component<ChildProps<InputProps, Re
         var queueLength = 0;
 
         var location = 'Unknown';
+        var taPhotoUrl = 'https://upload.wikimedia.org/wikipedia/commons/7/7c/Profile_avatar_placeholder_large.png';
```
Glitched: Same as above

server/index.ts
```diff
@@ -41,45 +53,79 @@ passport.use(new GoogleStrategy(
...
+            'users {' +
+            'userId' +
+            '}' +
+            '}' +
+            '}' +
+            '","variables":"' +
+            variablesString +
+            '"}';
+        request.post({
+            headers: { 'content-type': 'application/json' },
```
Glitched: Does the linter allow for a more sane representation of this query?

karunsingh: Mine doesn't unfortunately :(

client/src/components/includes/ConnectedSessionQuestions.tsx
karunsingh: Will do!

client/src/components/includes/AddQuestion.tsx
karunsingh: Good point, I'll change that! So then to keep things consistent, let's always use the same name as the GraphQL mutation name (might not be exactly the same as the mutation function)?

For eg
mutation xyz {
    apiXyz(...)
}
would give us <Mutation mutation={XYZ}>

PR #62, opened 2018-06-28
Refactor front end data flow

Glitched:
We're moving to two main data fetching queries. One for the SessionView (now a component) and one for the CalendarView (also now a component). The goal is to reduce redundancy and increase simplicity by limiting where we fetch data.

I deleted all the old types, which were mostly no longer in use in favor of new ones that match how we should be using the new schema. This had the side effect of requiring me to delete the old DetailedSessionsView so TS would be happy. After that, I deleted the popular questions stuff too, because at this point our structure has changed to the point where I don't think we'd save any time using that code. 

### Long Term Goals
(Outside the scope of this PR)
 * Unify HTML for Desktop/Mobile SessionInformationHeader and just use @media queries

Edit: This is currently in flux. AKA I'm messing around to see what works


client/src/components/includes/SessionQuestionsComponent.tsx
```diff
@@ -47,69 +39,46 @@ class SessionQuestionsComponent extends React.Component {
...
         const myQuestionCSS = this.props.isMyQuestion ? ' MyQuestion' : '';
 
         return (
             <div className={'QueueQuestions' + myQuestionCSS}>
                 {
                     this.props.isTA &&
                     <div className="studentInformation">
-                        <img src={this.props.studentPicture} />
-                        <span className="Name">{this.props.studentName}</span>
+                        <img src={question.userByAskerId.photoUrl} />
```
karunsingh: Is this checking for null somewhere upstream (and setting to `avatar`)?

client/src/components/includes/SessionQuestionsComponent.tsx
```diff
@@ -47,69 +39,46 @@ class SessionQuestionsComponent extends React.Component {
...
+                        </span>
                     </div>
                 }
-                <p className="Question">{this.props.studentQuestion}</p>
+                <p className="Question">{question.content}</p>
                 <div className="Tags">
-                    {tagsList}
+                    {question.questionTagsByQuestionId.nodes.map(
+                        (tag) => <SelectedTags
+                            key={tag.tagByTagId.name}
```
karunsingh: Although extremely unlikely, there's a possibility that two tags of a single question have the same name, in which case conflicting keys would mess things up. Setting the key to the `tagId` instead would avoid that!

client/src/components/includes/SessionQuestionsComponent.tsx
```diff
@@ -120,13 +89,13 @@ class SessionQuestionsComponent extends React.Component {
                                 <div className="TAButtons">
                                     <p
                                         className="Delete"
-                                        onClick={(e) => this._onClickDelete(e, updateQuestion)}
+                                        onClick={(e) => this._onClick(e, updateQuestion, 'noshow')}
```
karunsingh: I think we've been using `'no-show'` so far

client/src/components/includes/SessionQuestionsContainer.tsx
```diff
@@ -5,82 +5,55 @@ import { Icon } from 'semantic-ui-react';
...
-                                tags={questions[userQuestionIndex].tags}
-                                index={userQuestionIndex}
-                                isTA={this.props.isTA}
-                                isMyQuestion={true}
-                            />
-                        }
+                        <SessionQuestionsComponent
+                            key={myQuestion[0].questionId}
+                            question={myQuestion[0]}
+                            index={questions.indexOf(myQuestion[0])}
```
karunsingh: Hm I think we should rename this `index` prop in `SessionQuestionsComponent` to `position` or something -- I had to dive in to figure out what exactly it's doing

docs/authentication.md
```diff
@@ -25,7 +25,7 @@ We use Passport.js to handle our interaction with the Google sign-in service. Ou
...
 
-To store JWTs on the client-side, we use signed cookies. The package 'cookie-session' helps us with this, and in fact integrates seemlessly into the Passport flow. The Passport video tutorial series explains this well, but in essence, all we have to do is implement two functions: `serializeUser` and `deserializeUser`.
+To store JWTs on the client-side, we use signed cookies. The package 'cookie-session' helps us with this, and in fact integrates seemlessly into the Passport flow. The Passport video tutorial series explains this well, but in essence, all we have to do is implement two functions: `serializeUser` and `deserializeUser`. All cookies set should have both the `Secure` and `HttpOnly` flags set. The secure flag means the cookie will only be transmitted over a secure (HTTPS) connection, protecting from man-in-the-middle attacks. HttpOnly means that the cookies cannot be accessed in Javascript. This protects the user tokens from cross site scripting attacks (XSS).
```
karunsingh: I haven't tested it in production yet, but theoretically they should be served as secure + HttpOnly right now. (HttpOnly is true by default in the `cookie-session` package)

docs/setup.md
```diff
@@ -37,6 +37,8 @@ The catch here is that you still need the interaction with the server to work, s
 
 `> yarn start`
 
+Alternatively, if you are only doing front-end work, you may use the proxy key of the `client/package.json` and set it to `https://queue-me-in.herokuapp.com`. This will cause your development build to use the server implementation that we have hosted.
```
karunsingh: I'd add here that this backend might not be in sync with their current branch (for example, if someone else has deployed a more recent version of the server), so they should watch out for that

docs/setup.md
```diff
@@ -45,7 +47,7 @@ There is still one flaw though; our [authentication flow](./authentication.md) w
 
 `> yarn start --fakeuserid=<user-id-here>`
 
-For example, if we wanted to assume the identity of the user with ID 3 in our database, we could run `> yarn start fakeuserid=3`. This would artifically signal to the server that we are user 3, even though we're really not.
+For example, if we wanted to assume the identity of the user with ID 3 in our database, we could run `> yarn start --fakeuserid=3`. This would artifically signal to the server that we are user 3, even though we're really not.
```
karunsingh: Oops, good catch 👍

client/src/components/pages/SplitView.tsx
```diff
@@ -85,40 +60,6 @@ class SplitView extends React.Component {
...
-
-    // newDateIndex is an index between 0 and 6 inclusive, representing which of the days
-    // in the current week has been selected
-    handleDateClick(newDateIndex: number) {
-        this.setState({ selectedDateEpoch: this.state.selectedWeekEpoch + newDateIndex * ONE_DAY });
-    }
-
     // Keep track of active view for mobile
     handleSessionClick = (sessionId: number) => {
         this.props.history.push('/course/1/session/' + sessionId);
```
karunsingh: There's a couple of instances in this file where `/course/1` is hard-coded - we should plug in the `courseId` instead

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
+import * as React from 'react';
+
+import SessionInformationHeader from '../includes/SessionInformationHeader';
+import SessionQuestionsContainer from '../includes/SessionQuestionsContainer';
+
+import gql from 'graphql-tag';
+import { Query } from 'react-apollo';
+
+interface SessionData {
```
karunsingh: Let's shift this below `cost GET_SESSION_DATA`? It'd be more readable alongside `Variables` and `SessionDataQuery`, where it ends up getting used

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    userId
+                }
+            }
+        }
+    }
+    courseByCourseId(courseId: $courseId) {
+        name
+        code
+    }
+    sessionBySessionId(sessionId: $sessionId) {
```
karunsingh: Imagine this query with the old session series design 😨

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
...
+        isDesktop: boolean,
+        backCallback: Function,
+        joinCallback: Function,
+    };
+
+    render() {
+        return (
+            <section className="StudentSessionView">
+                {this.props.id === -1
+                    ? <p className="noSessionSelected">Please Select an Office Hour from the Calendar.</p>
```
karunsingh: Let's lowercase everything except the 'Please'

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    : <SessionDataQuery
+                        query={GET_SESSION_DATA}
+                        variables={{ sessionId: this.props.id, courseId: this.props.courseId }}
+                    >
+                        {({ loading, data, error }) => {
+                            if (error) { return <h1>ERROR</h1>; }
```
karunsingh: Add a TODO for error-handling

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
...
+                            if (!data || !data.sessionBySessionId) { return <div>Loading...</div>; }
```
karunsingh: We can change this to `if (loading)` I think

karunsingh: Eventually, we should make all the loading states look pretty with the spinners and centering etc

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
...
+    };
+
+    render() {
+        return (
+            <section className="StudentSessionView">
+                {this.props.id === -1
+                    ? <p className="noSessionSelected">Please Select an Office Hour from the Calendar.</p>
+                    : <SessionDataQuery
+                        query={GET_SESSION_DATA}
+                        variables={{ sessionId: this.props.id, courseId: this.props.courseId }}
```
karunsingh: Add a pollInterval here to refresh questions

client/src/components/includes/SessionView.tsx
```diff
@@ -0,0 +1,129 @@
...
+                                    <SessionInformationHeader
+                                        session={data.sessionBySessionId}
+                                        course={data.courseByCourseId}
+                                        callback={this.props.backCallback}
+                                        isDesktop={this.props.isDesktop}
+                                    />
+                                    <div className="splitQuestions">
+                                        <SessionQuestionsContainer
+                                            isTA={data.apiGetCurrentUser.nodes[0].
+                                                courseUsersByUserId.nodes[0].role === 'ta'}
```
karunsingh: the 'professor' role should also pass through with `isTA=true`

client/src/components/includes/CalendarView.tsx
```diff
@@ -0,0 +1,205 @@
...
+    apiGetSessions: {
+        nodes: [AppSession];
+    };
+    apiGetCurrentUser: CurrentUserRole;
+}
+
+const GET_CALENDAR_DATA = gql`
+    query getDataForDay(
+        $courseId: Int!,
+        $beginTime: Datetime,
```
karunsingh: We can make `$beginTime` and `$endTime` required as well

client/src/components/includes/CalendarView.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+import gql from 'graphql-tag';
+import { Query } from 'react-apollo';
+
+import CalendarHeader from '../includes/CalendarHeader';
+import CalendarDateSelect from '../includes/CalendarDateSelect';
+import CalendarSessions from '../includes/CalendarSessions';
+import CalendarWeekSelect from '../includes/CalendarWeekSelect';
+
+interface AppData {
```
karunsingh: Moving this below `GET_CALENDAR_DATA` would make it more readable (closer to where it's eventually used)

client/src/components/includes/CalendarView.tsx
```diff
@@ -0,0 +1,205 @@
...
+        selectedWeekEpoch: number,
+        selectedDateEpoch: number,
+    };
+
+    constructor(props: {}) {
+        super(props);
+        var week = new Date();
+        week.setHours(0, 0, 0, 0);
+
+        week.setDate(week.getDate() + 1 - week.getDay());
```
karunsingh: Hm, just occurred to me: could this mess up in those corner cases where it goes to a negative date?
For example, if it's Saturday the 1st, `week.getDate() = 1`, `week.getDay()=6` will evaluate to `-4` 🤔

karunsingh: Never mind, it seems to wrap around as expected

client/src/components/includes/CalendarView.tsx
```diff
@@ -0,0 +1,205 @@
...
+                    query={GET_CALENDAR_DATA}
+                    variables={{
+                        beginTime: selectedDate,
+                        endTime: new Date(this.state.selectedDateEpoch + ONE_DAY),
+                        courseId: this.props.courseId
+                    }}
+                >
+                    {({ loading, data, error }) => {
+                        if (error) { return <h1>ERROR</h1>; }
+                        if (!data) { return <div>Loading...</div>; }
```
karunsingh: `if (loading)` might work better

client/src/components/includes/CalendarView.tsx
```diff
@@ -0,0 +1,205 @@
...
+                        return (
+                            <aside className="CalendarView">
+                                <div className="Header">
+                                    <CalendarHeader
+                                        currentCourse="CS 3110"
```
karunsingh: `data.courseByCourseId.code`

client/src/components/includes/CalendarView.tsx
```diff
@@ -0,0 +1,205 @@
...
+                                        courseId={this.props.courseId}
+                                        isTa={data.apiGetCurrentUser && data.apiGetCurrentUser.nodes[0].
+                                            courseUsersByUserId.nodes[0].role === 'ta'}
```
karunsingh: or `=== 'professor'`

client/src/components/includes/CalendarSessionCard.tsx
```diff
@@ -30,52 +24,55 @@ class CalendarSessionCard extends React.Component {
...
+                            {/* Special class zero exists if we use the num ahead later */}
+                            <span className={'AheadNum '}>
+                                {session.questionsBySessionId.nodes.filter(q => q.status !== 'resolved').length}
+                            </span>
                         </span>
                         <span className="Finished">
                             Finished: &nbsp;
-                            <span className="FinishedNum">{this.props.resolvedNum}</span>
+                            <span className="FinishedNum">
+                                {session.questionsBySessionId.nodes.filter(q => q.status === 'resolved').length}
```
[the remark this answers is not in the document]
Glitched: I think we currently have the best behavior - but we can certainly discuss this

client/src/components/includes/CalendarView.tsx
Glitched: !data helps make TSX happy later on (so I can assume data exists) but now I changed it to loading || !data

client/src/components/includes/SessionInformationHeader.tsx
```diff
@@ -2,132 +2,50 @@ import * as React from 'react';
...
-                        }
+                        <p className="Location">{session.building + ' ' + session.room}</p>
+                        <Moment date={session.startTime} interval={0} format={'h:mm A'} />
+                        <Moment date={session.endTime} interval={0} format={' - h:mm A'} />
+                        <p className="Date">
+                            <Icon name="calendar" />
+                            <Moment date={session.startTime} interval={0} format={'dddd, D MMM'} />
+                        </p>
+                        <p>Held by <span className="black">
+                            {session.sessionTasBySessionId.nodes.map(ta =>
```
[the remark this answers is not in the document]
Glitched: In a word, poorly. But it should print out both names.

client/src/components/includes/SessionInformationHeader.tsx
```diff
@@ -138,51 +56,43 @@ class SessionInformationHeader extends React.Component<ChildProps<InputProps, Re
...
-                                        interval={0}
-                                        format={'dddd, D MMM'}
-                                    />
-                                }
+                                <Moment date={session.startTime} interval={0} format={'dddd, D MMM'} />
                             </p>
                         </div>
-                        <p>Held by <span className="black"> {tas.length > 0 && tas[0]} </span></p>
+                        <p>Held by <span className="black">
+                            {session.sessionTasBySessionId.nodes.map(ta =>
```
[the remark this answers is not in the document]
Glitched: Nope, this was incorrectly typed as non-nullable. Fixed.

client/src/components/includes/SessionView.tsx
Glitched: I can add loading, but because the others are nullable, TSX needs an option for those paths.

client/src/components/includes/SessionQuestionsContainer.tsx
Glitched: It's only used for onClick, which is only used in AddQuestion to my knowledge

client/src/components/includes/CalendarSessionCard.tsx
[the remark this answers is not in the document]
karunsingh: Yup, I think this makes sense! Just something to watch out for if it's ever a source of confusion, but I believe this is the most intuitive definition of 'Finished'

client/src/components/includes/SessionView.tsx
karunsingh: 😄

karunsingh:
Looks good! Press the button whenever you're ready 😃

PR #63, opened 2018-07-02
Authentication Patch

karunsingh:
- Fixes the authentication flaw in #61 by only allowing the server to call `api_find_or_create_user`
- Adds a database trigger to the courses table, which adds every existing user as a student whenever a new course is inserted
- Updated database documentation with these changes

### DIFF
--- a/server/office_hours.sql
+++ b/server/office_hours.sql
@@ -360,12 +360,17 @@ CREATE FUNCTION public.api_find_or_create_user(_email text, _google_id text, _fi
 
 declare
 
+caller integer;
 _user_id integer;
 course_row courses%rowtype;
 
 begin
+	caller := (current_setting('jwt.claims.userId', true)::integer);
+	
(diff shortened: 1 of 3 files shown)


server/office_hours.sql
```diff
@@ -539,6 +544,36 @@ end
...
+
+begin
+
+	for user_row in
+
+		select * from users
+
+	loop
+
+		insert into course_users (course_id, user_id, role) values (new.course_id, user_row.user_id, 'student');
```
Glitched: I'm not sure if this is where it happens, but it seems although I can't login, I can still create users by hitting `apiFindOrCreateUser` directly from GraphQL.  This is better than where we were, but by using a bad user id, you can still lock people out from creating accounts by using their email. (because both email & user ID must be unique)

Glitched:
LGTM!

PR #66, opened 2018-07-15
Frontend Fixes

Glitched:
Done:
=====
* Refactored AddQuestion to reduce props & use state machine for logic
* Restyled ‘NOW’ indicator in session questions component
* Repurpose ProfessorHeader as TopBar and use it in SessionView
* Tweaked Login Size behavior 
* ‘Welcome, xyz’ when no office hour is selected in SplitView
* Handling cases with 2 TAs
* Move to computedAvater & computedName
* Add polling to data queries
* Fix backend so it can run on Heroku
* Fix addQuestion back button (push state)
* Swap mobile course name for course code


client/src/components/App.tsx
```diff
@@ -58,6 +58,7 @@ class App extends React.Component {
                         <Link to="/course/1"> Split View</Link>
                     </nav>
                     <Switch>
+                        <Redirect from="/__auth" to="/course/1" />
```
karunsingh: Hmm, is this implementing the redirect after login? I think we should just try redirecting `/` to `/course/1` -- `/__auth` is the URL that the login button links to I believe

client/src/components/pages/ProfessorView.tsx
```diff
@@ -12,11 +12,13 @@ class ProfessorView extends React.Component {
...
                 />
-                <div className="rightOfSidebar">
-                    <ProfessorHeader
-                        professor="Michael Clarkson"
-                        image="https://www.cs.cornell.edu/~clarkson/img/mrc_gates300.jpg"
-                        notification={true}
+                <section className="rightOfSidebar">
+                    <TopBar
+                        user={{
+                            computedName: 'Michael Clarkson',
```
karunsingh: I think this was the other minor thing we couldn't remember today; we can change this to also reflect the current logged-in user. We can probably factor that data out into the top bar (which will soon include the calendar top bar as well)

client/src/components/includes/AddQuestion.tsx
```diff
@@ -9,320 +9,188 @@ import SelectedTags from '../includes/SelectedTags';
...
+        }) {
         clientMutationId
     }
 }
 `;
 
 class AddQuestion extends React.Component {
-
+    /*
+     * State machine states
```
karunsingh: 💯

client/src/components/includes/AddQuestion.tsx
```diff
@@ -9,320 +9,188 @@ import SelectedTags from '../includes/SelectedTags';
...
-                                value={this.state.question}
-                                onChange={this.handleClick}
-                                placeholder="What's your question about?"
-                            />
-                            : <p className="placeHolder text">Finish selecting tags...</p>}
-                    </div>
-                    <Mutation mutation={ADD_QUESTION}>
-                        {(addQuestion) =>
-                            this.state.doneSelectingTags ?
+                                {this.state.stage <= 10 ?
```
karunsingh: I think the only thing that got lost in this part of the refactor is the 'activated' boolean property on primary tags

client/src/components/includes/AddQuestion.tsx
```diff
@@ -9,320 +9,188 @@ import SelectedTags from '../includes/SelectedTags';
...
+                                    this.props.tags
+                                        .filter((tag) => tag.level === 1)
```
karunsingh: adding `.filter((tag) => tag.activated)` should do it, provided that the activated property is being passed down

client/src/components/pages/ProfessorView.tsx
Glitched: I think this might be better for another PR, where we're already getting data from the backend. In this one, we aren't getting any data from the server yet.

client/src/components/App.tsx
Glitched: It was working locally on my machine 0.o, but I changed it to this so it should work in prod.

client/src/components/includes/AddQuestion.tsx
Glitched: Those tags are server tag objects, not selected tags. They don't have any state. That said, I found a simpler way to do the filtering

client/src/components/includes/AddQuestion.tsx
Glitched: I keep a list of activated tag ids which should suffice as a proxy for that, without all the overhead of maintaining that.

client/src/components/includes/AddQuestion.tsx
karunsingh: Hmm by activated I meant the database property for each tag that indicates whether or not it is currently active (i.e. visible) for students in the course. We just need to make the primary tags (+ their children) not available for selection if they're not active in the DB

PR #67, opened 2018-07-26
Undo Feature

karunsingh:
- Implements the undo feature (front-end and backend) for both, TAs and students

Things to discuss:
- Queue updating lag is very noticeable at 4 seconds when the user is taking some action and expects an instant response (such as add question or undo)
- Undo message disappears after 15 seconds (currently arbitrary - should it be less or more?)


client/src/components/includes/SessionView.tsx
```diff
@@ -86,43 +99,139 @@ class SessionView extends React.Component {
...
+        super(props);
+        this.state = {
+            undoAction: undefined,
+            undoName: undefined,
+            undoQuestionId: undefined,
+            timeoutId: null,
+        };
+        this.triggerUndo = this.triggerUndo.bind(this);
+        this.dismissUndo = this.dismissUndo.bind(this);
+        this.handleUndoClick = this.handleUndoClick.bind(this);
```
Glitched: In my PR, I removed all the bind() lines by using arrow functions, would you mind quickly doing that? 
There's no real reason we have to do this for every function in a constructor.

Glitched:
Looks great!

PR #68, opened 2018-07-28
Logout

Glitched:
* Logout functionality for mobile and desktop
* Switch views for profs
* Restyle calendar
* Share feedback form (only mobile rn)
* Fix placeholder in computedAvatar
* Call refetch when data is modified
* Make AddQuestion height perfectly consistent for all states

Notes:
I briefly tried to update to React 16.4 and TS 2.9 in this PR to use the fancy new refs, but that looks like it will be a lot of work. 

The big (wonderful) change is moving from `Function` to actual types i.e `(string) => bool`. That won't be terribly bad, but I encountered some weird TS bugs that can be kicked down the road.
### Todos:
* Missing icons


client/src/components/includes/CalendarDaySelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+        };
+    }
+
+    incrementWeek = (forward: boolean) => {
+        if (forward) {
+            this.setState({ selectedWeekEpoch: this.state.selectedWeekEpoch - ONE_WEEK });
+        } else {
+            this.setState({ selectedWeekEpoch: this.state.selectedWeekEpoch + ONE_WEEK });
+        }
+    }
```
karunsingh: At the end of this function, we should propagate the new state to CalendarView so that it refetches sessions:
`this.props.callback(this.state.selectedWeekEpoch + this.state.active * ONE_DAY);` should do it!

PR #71, opened 2018-08-11
Fix link to docs

SheaBelsky:
Link was broken before. Relative linking in GitHub is weird. Changed to absolute link.


karunsingh:
Thanks!

PR #72, opened 2018-08-12
Jw/add question check duplicate

jillwuu:
Added a check to see if asked has asked other questions that are currently unresolved, and if so, if the questions are in non-expired sessions. Throws an exception if such a condition is found to be true.

### DIFF
--- a/server/office_hours.sql
+++ b/server/office_hours.sql
@@ -84,8 +84,17 @@ inserted_question integer;
 tag integer;
 asker users%rowtype;
 _asker_id integer;
+checked_session_ids integer[];
+questions questions%rowtype;
 begin
 	select * into asker from api_get_current_user();
+	select count(*) > 0 into question_asked from questions where asker_id = _asker_id AND status = 'unresolved';
+	if (question_asked > 0) then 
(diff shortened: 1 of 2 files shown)


server/office_hours.sql
```diff
@@ -84,8 +84,25 @@ inserted_question integer;
 tag integer;
 asker users%rowtype;
 _asker_id integer;
+checked_session_ids integer[];
+end_time timestamp without time zone;
+questions questions%rowtype;
 begin
 	select * into asker from api_get_current_user();
+	select count(*) > 0 into question_asked from questions where asker_id = _asker_id AND status = 'unresolved';
```
karunsingh: Just realized that `_asker_id` is undefined at this point; could we move this entire block to the `else` block below? (Right after we define `_asker_id`

server/office_hours.sql
```diff
@@ -84,8 +84,25 @@ inserted_question integer;
...
+	permission = true;
```
karunsingh: Sorry I didn't catch this before, but seems like `permission` is unused so we can remove it 👍

PR #104, opened 2018-12-05
Professor Analytics Graphs

rsn55:
### Inside This PR

* Creates a new tab, "People", in the professor view, which has 3 graphs showing various office hour questions stats based on a user-selected date.

### Notes 

* Nivo currently does not offer a custom tooltip for the line graph, but this will probably be released soon. For now, the tooltip is just turned off.

### Breaking Changes

### Checklist:

- [x] My PR adds a @ts-ignore
- [x] I resolved any merge conflicts


client/package-lock.json
```diff
@@ -0,0 +1,16042 @@
+{
+    "name": "client",
+    "version": "0.1.0",
```
Glitched: We won't be able to deploy this branch if it has a package-lock.json file because it conflicts with Yarn's lock - could you delete this?

client/package.json
```diff
@@ -4,6 +4,10 @@
     "private": true,
     "proxy": "http://localhost:3001",
     "dependencies": {
+        "@nivo/bar": "^0.51.5",
+        "@nivo/line": "^0.51.0",
+        "@nivo/pie": "^0.51.0",
+        "@types/react-dates": "^17.1.3",
```
Glitched: Could you move this @types line down to dev dependencies with the other types?

Typescript doesn't actually do anything in production - it's just checking types while we work. Dev dependencies (like @types packages) are ones only used in development

client/src/components/includes/ProfessorSidebar.tsx
```diff
@@ -9,7 +9,7 @@ class ProfessorSidebar extends React.Component {
     };
 
     render() {
-        var selectedArray: string[] = ['', '', ''];
+        var selectedArray: string[] = ['', '', '', ''];
```
Glitched: If you're still looking for things to refactor, this selectedArray is pretty ugly. It seems like we could do a selected index instead. Maybe even express the items in the sidebar as a list of objects {name, icon, route} and just map over that to the same JSX we repeat a few times.

If you don't need more things, then you can leave this as is.

githubKey
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
Glitched: Uhh, I don't know what this is the key to, but this is probably not good

package.json
```diff
@@ -19,7 +19,16 @@
     },
     "homepage": "https://github.com/cornell-dti/office-hours#readme",
     "dependencies": {
+        "@nivo/line": "^0.50.0",
```
Glitched: Could you revert the changes in this file? I think they might have been on accident

client/src/components/includes/QuestionsBarChart.tsx
```diff
@@ -0,0 +1,156 @@
...
+        if (!(this.isEmpty(this.props.sessionDict))) {
+
+            var ta = this.props.sessionDict[sessionId].ta;
+            var start = this.props.sessionDict[sessionId].startHour;
+            var end = this.props.sessionDict[sessionId].endHour;
+            var answered = this.props.sessionDict[sessionId].answered;
+            var questions = this.props.sessionDict[sessionId].questions;
+            var percent = Math.round((answered / (questions)) * 100);
+            var building = this.props.sessionDict[sessionId].building;
+            var room = this.props.sessionDict[sessionId].room;
```
Glitched: Instead of pulling all these variables out, could you do something like:
`var session= this.props.sessionDict[sessionId];`
then just use `session.ta` etc?

PR #108, opened 2019-02-04
Sa/course switcher

Shefali-A:
Finished implementing the course switcher.

### Inside This PR

*

### Breaking Changes

### Checklist:

- [ ] My PR adds a @ts-ignore


client/src/components/includes/CalendarHeader.tsx
```diff
@@ -10,32 +10,39 @@ class CalendarHeader extends React.Component {
...
     state: {
         showMenu: boolean;
+        showCourses: boolean;
     };
 
     constructor(props: {}) {
         super(props);
-        this.state = { showMenu: false };
+        this.state = { showMenu: false, showCourses: false };
+        // this.state = { showCourses: false };
```
Glitched: Can you delete this instead of leaving it commented?

client/src/components/includes/CalendarHeader.tsx
```diff
@@ -47,21 +54,32 @@ class CalendarHeader extends React.Component {
...
                             </React.Fragment>
-                        } */}
+                        } }
                         <li><a href="/__auth/logout"><span><Icon name="sign out" /></span>Log Out</a></li>
                         <li><a href="https://goo.gl/forms/7ozmsHfXYWNs8Y2i1" target="_blank">
                             <span><Icon name="edit" /></span>Send Feedback</a>
                         </li>
                     </ul>
-                )}
+                )*/}
```
Glitched: We can't just comment this out - it's the logout menu on mobile

Glitched:
Hey Shefali, it seems you deleted the `client/package.json` instead of `package-lock.json`. Mind fixing this?

We need package.json to run the client on our machine and package-lock.json is a file that nom creates that breaks deploying to production

Shefali-A:
> Hey, there's a couple things we need to take care of before we can merge this in.
> In addition to the comments I made, you need to delete `package-lock.json` (we can't deploy while it's here) and undelete `yarn.lock`
> Also, could you pull master into this branch? I can't run it locally due to some changes we've had to make

Glitched:
Hey, so the bug that prevented it from launching was super simple - you just didn't uncomment correctly. If you don't mind, please verify your changes before you commit them up. 

Also going forward I noticed a couple of things

- The Chevron was still commented out, so users wouldn't be able to find the dropdown
- You reused the .logoutMenu CSS, but changed it so that it broke the logoutMenu. Please watch out for stuff like this
- It was showing all classes in the database, so I had to limit them to one semester. I had to add a little tech debt to get this out the door by hardcoding the current semester in CalendarHeader. We'll need to move this to the DB at some point.
- The tap targets for the link was only the text, which was a little confusing because the menu was disappearing on me, but I wasn't being taken anywhere.

PR #153, opened 2019-10-13
Firebase Collection Update Upon Log In

jordepic:
### Inside This PR 

* Upload User's first name, last name, account creation time, photo url, email, and last time signed in onto users/(current user uid)/ in the firebase collection.
* This occurs every time that a user either signs in manually with the popup or automatically by Google.

### Test Plan 
* Only test on queue-me-in staging.

### Notes 

* When a user is automatically signed in by Google (and they never have to click on their account in the popup window), their last sign in is still considered to be whenever they last used the popup window.  If this time were to be changed to account every single time that they were logged in, it would require switching the time format in the database (which is not hard).

### Breaking Changes 

### Checklist

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/client/src/components/pages/LoginView.tsx
+++ b/client/src/components/pages/LoginView.tsx
@@ -3,8 +3,9 @@ import { Icon } from 'semantic-ui-react';
 const QMeLogo = require('../../media/QLogo2.svg');
 const googleLogo = require('../../media/googleLogo.svg');
 
-import firebase, { app } from '../../firebase';
+import firebase, { app, firestore } from '../../firebase';
 import { useHistory } from 'react-router-dom';
+import { userUpload } from '../../firehooks';
 
 const LoginView: React.FC = () => {
     const [showContact, setShowContact] = React.useState(false);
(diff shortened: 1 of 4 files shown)


Glitched:
This is a good effort!
A couple of minor tweaks here, in general we try to avoid using ‘var’ to declare variables. We prefer to use ‘const’ if the value isn’t updated after it is declared or ‘let’ if it is. 
https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75

I think we should break your userUpload function into a separate file. Firehooks contains a bunch of functions that are used as React Hooks and I think maintaining a separation here would be a good idea. Maybe just break it into its own file for now and we’ll see if we can group it logically later? 

One final note that I don’t have a solution for: right now this depends on us trusting all the data coming from the user. As in, I could call your function myself and upload whatever I want to the users table. We should have some discussion as to what we can verify. 

I’m pretty sure a security policy can allow people to only write /users/uid. We’d also need to make sure that the email address is correct at the very least because that’s how we tie students/professors to real life. 

I suppose there’s no harm in the rest

client/src/components/pages/LoginView.tsx
```diff
@@ -21,8 +22,9 @@ const LoginView: React.FC = () => {
         return app
             .auth()
             .signInWithPopup(authProvider)
-            .then((response: {}) => {
-                // RYAN_TODO Create or update user
+            .then((response) => {
+                var user = response.user;
```
BobbyVillaluz: Could we change this to `const user = response.user;`?

PR #156, opened 2019-10-20
Leave Other Queues

connielei:
### Inside This PR 
* Allows user to click on the text that says "You are holding a spot in another active queue...." to remove their question in the other active queue. After, the join the queue button will be clickable and the other text will disappear.

### Test Plan
* Tested locally by adding joining a class's queue and then changing classes to get the "You are holding..." text to display, then pressed the text and added a question to the queue.
* Need a way for testing what other uses see when someone removes their question to join a different queue.

### Notes
* A list of question ids are passed into SessionQuestionsContainer, but we only remove the first index/first question id. There shouldn't be a case where the list has length greater than one, but if that ever comes up, we would need to either add a database call that updates the statuses given multiple question ids or loop through the list and use the single call multiple times.

### Breaking Changes 
- [X] I updated existing types in `/src/components/types/`

### Checklist
- [ ] My PR adds a @ts-ignore

### DIFF
--- a/client/src/components/includes/SessionQuestionsContainer.tsx
+++ b/client/src/components/includes/SessionQuestionsContainer.tsx
@@ -102,105 +102,101 @@ class SessionQuestionsContainer extends React.Component {
         // console.log('questions associated w/ this session', questions);
 
         return (
-            <div className="SessionQuestionsContainer splitQuestions" >
-                {!this.props.isTA && myQuestion && myQuestion.length === 0 && this.props.isOpen
-                    && !this.props.haveAnotherQuestion &&
-                    <div className="SessionJoinButton" onClick={() => this.props.handleJoinClick()}>
-                        <p><Icon name="plus" /> Join the Queue</p>
-                    </div>
-                }
+            <React.Fragment>
                 {!this.props.isTA && myQuestion && myQuestion.length === 0 && this.props.isOpen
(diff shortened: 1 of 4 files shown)


client/src/components/includes/SessionQuestionsContainer.tsx
```diff
@@ -74,90 +99,104 @@ class SessionQuestionsContainer extends React.Component {
             }
         }
 
+        // console.log('questions associated w/ this session', questions);
```
BobbyVillaluz: Before pushing to master it's a good practice to delete the console.log's that we used for debugging.

client/src/styles/SplitView.less
```diff
@@ -23,6 +23,20 @@ section.StudentSessionView {
...
+.SessionLeaveQueueMessage{
+    color: #ffffff;
+    background-color:#8d8d8d;
+    font-size: 16px;
+    padding: 12px 0px 12px 0px;
+    font-weight: 100;
+    margin-top: -13px; //this is - 13px, not -12px because it is fuzzy
+}
+
+.removeQuestionLeaveQueue{
```
BobbyVillaluz: Adding a `cursor: pointer;` to this class will change the cursor to a hand when the user hovers over it.

connielei: Why do we want the white hand pointer over an arrow?

PR #178, opened 2019-11-05
Protect professor routes on firebase

Glitched:
Get the courseUser value before we show info to professors

### Inside This PR 

* That's all, folks

### Test Plan 
Manually verified changes

### Notes 

Might be cool to pass the courseUser we get here as a prop into profviews to eliminate duplicate data (if we use that somewhere).

### Breaking Changes 

### Checklist

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/client/src/components/App.tsx
+++ b/client/src/components/App.tsx
@@ -39,10 +39,10 @@ const PrivateRoute = ({ component, requireProfessor, ...rest }: any) => {
     // 1: Not logged in
     // 2: Logged in
 
-    const courseId = requireProfessor ? rest.computedMatch.params.courseId : 'JUNK VALUE';
+    const courseId = requireProfessor ? rest.computedMatch.params.courseId : undefined;
 
     const [isLoggedIn, setIsLoggedIn] = React.useState(0);
-    const courseUser = requireProfessor && useMyCourseUser(courseId);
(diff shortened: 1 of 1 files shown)


client/src/components/App.tsx
```diff
@@ -48,9 +53,11 @@ const PrivateRoute = ({ component, ...rest }: any) => {
         }
     });
 
-    if (isLoggedIn === 0) {
+    if (isLoggedIn === 0 || (requireProfessor && !courseUser)) {
```
BobbyVillaluz: Can you explain what this second condition is checking?

Glitched: This if statement is to show the loader. There's two conditions to show the loader
1. We haven't loaded the user info yet
2. We have a route that requires professor permissions and we don't know if the user is a professor.

This new condition takes requireProfessor, a new prop I added to our `PrivateRoute` component to know whether to verify the user is a professor before rendering. The `!courseUser` is a shorthand for checking if courseUser is undefined, i.e. has not finished fetching data. 

`!{} === false` and `!undefined === true`

This logic used to be split across each professor page, but this PR centralizes the feature where it makes more sense. We used to have it in the page because that's where we fetched all our data

BobbyVillaluz: Great, thanks for the explanation!

PR #191, opened 2019-12-09
Migrate from create-react-app-typescript to official CRA

Glitched:
Replace create-react-app-typescript with the official create-react-app. Create-react-app-typescript is deprecated and no longer receiving updates in favor of official typescript support. 

### Inside This PR

* Replace react-scripts-ts with react-scripts
* Update code to work with stricter linter

### Notes 

Moving away is essential for the long term health of the project and I figured I would do it was I was here because I've spent more time fighting with Typescript than most and some of the changes were non-obvious.

I fixed all the breaking errors, but this created a bunch of warnings (where we were doing bad stuff). We should probably get around to fixing all of them at some point, but I think an incremental adoption strategy would work best. 

Future work involves moving from the deprecated TSLint to ESLint. Then we should improve our linting for accessibility, code format, and code safety.
### Breaking Changes

- [x] Other change that could cause problems (Detailed in notes)

### Checklist:

- [ ] My PR adds a @ts-ignore
- [ ] I tested affected functionality
- [ ] My PR is ready for review

### DIFF
--- a/client/src/components/App.tsx
+++ b/client/src/components/App.tsx
@@ -15,7 +15,7 @@ import ProfessorPeopleView from './pages/ProfessorPeopleView';
 import { Analytics } from './includes/Analytics';
 import { Loader } from 'semantic-ui-react';
 import { userUpload } from '../firebasefunctions';
-import { useMyCourseUser } from 'src/firehooks';
+import { useMyCourseUser } from '../firehooks';
 
 ReactGA.initialize('UA-123790900-1');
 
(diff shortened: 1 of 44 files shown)


client/src/components/App.tsx
```diff
@@ -15,7 +15,7 @@ import ProfessorPeopleView from './pages/ProfessorPeopleView';
 import { Analytics } from './includes/Analytics';
 import { Loader } from 'semantic-ui-react';
 import { userUpload } from '../firebasefunctions';
-import { useMyCourseUser } from 'src/firehooks';
+import { useMyCourseUser } from '../firehooks';
```
SamChou19815: Latest CRA can actually support it. You can set the `baseUrl` in tsconfig. Not push blocking.

client/src/components/includes/Analytics.tsx
```diff
@@ -11,7 +11,7 @@ interface AnalyticsProps {
 }
 
 export class Analytics extends React.Component {
-    props: AnalyticsProps;
+    props!: AnalyticsProps;
```
SamChou19815: Probably
```tsx
export class Analytics extends React.Component<AnalyticsProps> {}
```

Glitched: Makes sense. I just did this to get it to compile. If I have time, I'd love to make it a hooks component

PR #207, opened 2020-02-23
Support changing course user roles

SamChou19815:
Support changing the course user roles in the professor's manage user view.

### Inside This PR

1. Added the id of the course user document to the `enrichedFireCourseUser` object.
2. Use the id of the course user to perform role change update
3. Removed RYAN_TODO for loading and error. With firebase, it should be handled by the parent component, and the dropdown component should just follow its props.

### Notes 

This diff does come with a but: The functionality does not completely work, most likely because the security rules are not properly setup. I strongly suspect that it's caused by firebase's poor and undocumented support for references in security rules. I might experimenting changing all references to strings in a future diff.

However, the stuff does work if I disable the security rules or changing the type of the course id and user id to string.

### Breaking Changes

### Checklist:

- [x] NEW: I removed a `RYAN_TODO`
- [ ] My PR adds a @ts-ignore
- [ ] I tested affected functionality
- [ ] My PR is ready for review

### DIFF
--- a/client/src/components/includes/ProfessorRolesTable.tsx
+++ b/client/src/components/includes/ProfessorRolesTable.tsx
@@ -5,32 +5,32 @@ import * as _ from 'lodash';
 
 import { firestore, collectionData } from '../../firebase';
 import { switchMap, map } from 'rxjs/operators';
-import { docData } from 'rxfire/firestore';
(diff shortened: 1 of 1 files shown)


SamChou19815:
Loops in @Glitched about the document reference issue.

PR #245, opened 2020-03-08
Queue Fixes

abcdefguan:
Begin implementation of questions

### Inside This PR

* Added semester and year fields to courses
* Fixed bug where questions in queue did not appear in ascending order of timestamp
Fixed in the backend => Acquired questions by ascending order of timestamp
* Added field "resolved" to questions
=> Firebase does not support != queries, so a "resolved" field with boolean value is needed
* Disabled tags in security rules and timeAddressed because there was some bug with the tags (idk why). timeAddressed is not always present.
* Added "assigned" as valid status in security rules
* Fixed bug where question answerers, tags, asker etc do not real time update correctly
=> Added state, getDerivedStateFromProps and some caching to fetch answerers correctly
=> answers, tags, asker will no longer real time update with changes in their information (ie if Connie changes her name to Jenny, Connie will no longer be updated to Jenny)

### Notes 

Introduced new bug:
SessionQuestion.tsx is using about twice the amount of gets that it should be using. I believe that this is an issue with me not understanding the React lifecycle. It could have something to do with Firebase and the fact that it returns on a query twice.

Implemented:
* Assign question
* No show for question
* Done for question
* Questions that are resolved no longer show up on queue

Todo:
* Automatic assign for the first person in queue (done with a transaction)
* Shift owned questions up
* Handle Undos
* Unassign
* Skip function

Questions
* Was undo implemented successfully?
* Why is there no don’t know button?

### Breaking Changes

- [x] Database schema change (anything that changes Postgres)
- [x] I updated existing types in `/src/components/types/`

* Added "resolved" field to questions
=> Firebase does not support != queries, so a "resolved" field with boolean value is needed
* Changed security rules by disabling tag checks (some bug here) and timeAssigned (not always present). Added "assigned" as valid status in security rules
* Added "semester" and "year" fields to courses

### Checklist:

- [x] My code follows the code style of this project. (Why do you require semicolons? Why?)
- [x] My changes requires a change to the documentation.
- [ ] My PR adds a @ts-ignore


client/src/components/includes/SessionQuestion.tsx
```diff
@@ -67,27 +81,149 @@ class SessionQuestion extends React.Component<Props> {
...
+            stateChanges.loading = false;
+        };
+
+        if (state.loadedAskerId !== props.question.askerId){
+            stateChanges.loadedAskerId = props.question.askerId;
+            stateChanges.asker = null;
+            stateChanges.loading = false;
         };
 
+        return Object.assign(state, stateChanges);
```
SamChou19815: Use `{...state, ...stateChanges}` instead

client/src/components/includes/SessionQuestion.tsx
```diff
@@ -67,27 +81,149 @@ class SessionQuestion extends React.Component<Props> {
...
+        //Primary and secondary tags
         const primaryTag$: Observable<FireTag>
             = docData(firestore.doc('tags/' + props.question.primaryTag), 'tagId');
         primaryTag$.subscribe(primaryTag => this.setState({ primaryTag }));
 
         const secondaryTag$: Observable<FireTag>
             = docData(firestore.doc('tags/' + props.question.secondaryTag), 'tagId');
         secondaryTag$.subscribe(secondaryTag => this.setState({ secondaryTag }));
-    }
+    }*/
```
SamChou19815: No commented code please

client/src/components/includes/SessionQuestion.tsx
```diff
@@ -123,6 +260,34 @@ class SessionQuestion extends React.Component<Props> {
         });
     };
 
+    //This function attempts to assign this question to the current user
```
SamChou19815: This line of docs seems to just paraphrase the function name. Same problem for two below.

client/src/components/includes/SessionQuestion.tsx
```diff
@@ -269,12 +437,19 @@ class SessionQuestion extends React.Component<Props> {
                                 <React.Fragment>
                                     <p
                                         className="Delete"
-                                    // onClick={(e) => this._onClick(e, updateQuestion, 'no-show')}
```
SamChou19815: No commented code

server/package-lock.json
```diff
@@ -0,0 +1,14 @@
...
+    "requires": true,
+    "dependencies": {
+        "@types/node": {
+            "version": "10.17.17",
+            "resolved": "https://registry.npmjs.org/@types/node/-/node-10.17.17.tgz",
+            "integrity": "sha512-gpNnRnZP3VWzzj5k3qrpRC6Rk3H/uclhAVo1aIvwzK5p5cOrs9yEyQ8H/HBsBY0u5rrWxXEiVPQ0dEB6pkjE8Q==",
+            "dev": true
+        }
+    }
+}
```
SamChou19815: This file should not be there.

client/src/components/includes/SessionQuestion.tsx
```diff
@@ -67,26 +81,123 @@ class SessionQuestion extends React.Component<Props> {
...
+                (doc) => {
+                    if (doc.exists){
+                        this.setState({
+                            secondaryTag: doc.data()
+                        });
+                    }
+                }
+            );
+            shouldUpdateLoad = true;
+        }
```
SamChou19815: This is a little concerning. We are paying `O(n)` IO cost of fetching tags for all questions.
I think a better way to deal with this is to:
- Fetch all tags belong to a class inside the parent component (which shouldn't be too many)
- For each tag Id, find the tag with that id, and pass it into `SessionQuestion`.

client/src/components/includes/SessionQuestion.tsx
```diff
@@ -67,26 +81,123 @@ class SessionQuestion extends React.Component<Props> {
...
+                (doc) => {
+                    if (doc.exists){
+                        this.setState({
+                            answerer: doc.data()
+                        });
+                    }
+                }
+            );
+            shouldUpdateLoad = true;
+        };
```
SamChou19815: This is the tricky part. You cannot efficiently prefetch all users in a class, because there is an indirection of course user -> users. It might be worth thinking whether we should merge these two tables to allow more efficient querying of questions.

client/src/components/includes/SessionQuestion.tsx
abcdefguan: The costs were O(n) before I made the changes, but I guess it's more expensive on firebase than with a SQL server. I think I'm getting somewhat better results by memoizing the thing.
I agree with your suggestion of caching the tags on the parent then passing them as props to the children (essentially doing firebase work on client side). I'll work on that on the next fix (which might be next week).
Could you approve the pull request please? Thanks.

client/src/components/includes/SessionQuestion.tsx
abcdefguan: I think we might want to discuss this in person or something. I think your optimization will be a tradeoff between server cost and development time and might require discussion with the team. Since we still haven't implemented many features yet, it might be wise to spend development time on features rather than optimizations for now.

PR #277, opened 2020-03-27
Reorganize directories

SamChou19815:
### Summary

In `develop-firebase` branch, we completely eliminates the old backend and there is no need for the `server` folder any more! Therefore, I pulled everything from `client` to the root. As a result, some CI configurations are tweaked to reflect this.

### Test Plan

CI checks passing. `yarn start` still works.

### Checklist

- [ ] My PR adds a @ts-ignore


SamChou19815:
> [diff-counting] Significant lines: 2190. This diff might be too big! Developer leads are invited to review the code.

lol. The autogenerated README deletion is picked up by the bot.

PR #285, opened 2020-04-02
Routing logic improvements

SamChou19815:
### Summary

Right now, our routing logic in `App.tsx` is a complete mess. The condition check for login status, data loading status, permissions, and course redirect logic are merged together in a giant block. Sometimes, the order of those checks matters and sometimes it doesn't. It's not immediatelty clear from this giant mess what's the correct flow of check. What's worse, it's also hard to reason about whether the current logic is correct. (hint: it's not)

This diff aims to clean up the giant mess by breaking it down into layers of composible react hooks. Each layer checks one thing only. The routing flow is therefore easier to understand: you unwrap the hooks layer by layer to see what's necessary to check in each step.

This change allows us to add a default route to the end of the routing switch without a lot of copy-pasting code. It fixed the issue when visiting `/` at the start already redirect you to `/edit`. (This is the incorrect behavior I mentioned earlier.)

### Test Plan

Added test plan to the top of the App.tsx

### Checklist

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,3 +1,13 @@
+/**
+ * If you changed the routing logic, make sure to test the following stuff to ensure the behavior
+ * is correct:
+ *
+ * 1. Try sign out and visit private endpoints
+ * 2. Try sign in as student and visit an unauthorized professor course.
+ * 3. Try sign in as student with zero courses
+ * 4. Try sign in as student with some courses
+ */
+
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


SamChou19815:
> If a student tries to access professor view, the student gets logged out and sent to the login screen. Is this intended?

This is the old behavior before I touched any code. Ideally, there should be a permission denied page.

PR #312, opened 2020-04-09
Display all questions

### LINKED ISSUE #295: UI Issue when submitting to queue with >=10 questions
When submitting a question to a queue with >=10 questions, the user's submitted question does not appear under "my questions". The UI also does not reflect the user's question. Since the one question per queue limit is enforced on the client side, this bug allows the user to submit multiple questions if there are currently >=10 questions on the queue.

SamChou19815:
### Summary

With firebase realtime listening, we no longer need to limit the number of questions to reduce server-side load. With this change, we can fix #295.

### Test Plan

Add more than 10 questions, no longer have the problem mentioned in #295.

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -253,18 +253,20 @@ class SessionQuestion extends React.Component<Props, State> {
                     <div className="Location">
                         {this.props.isTA && question.location}
                     </div>
-                    {(this.props.isTA || includeBookmark || this.props.includeRemove) &&
-                        <p className={'Question' + studentCSS}>{question.content}</p>}
+                    {(this.props.isTA || includeBookmark || this.props.includeRemove) ?
+                        <p className={'Question' + studentCSS}>{question.content}</p> :
+                        <p className={'Question' + studentCSS}>Another Student's Question</p>}
                     {question.taComment && (
                         <p className={'Question' + studentCSS}>TA Comment: {question.taComment}</p>
                     )}
                 </div>
                 <div className="BottomBar">
                     {this.props.isTA && <span className="Spacer" />}
+                    {(this.props.isTA || includeBookmark || this.props.includeRemove) &&
                     <div className="Tags">
                         {primaryTag && <SelectedTags tag={primaryTag} isSelected={false} />}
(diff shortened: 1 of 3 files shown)


SamChou19815:
> Prevented students from seeing tags of other students' questions (potential security vulnerability)
> Wrote "Another Student's Question" on questions that you cannot see to clarify to end user

We want to probably revert it to preserve the original behavior. We should not proceed without design.

abcdefguan:
I think it's fine to revert the "Another Student's Question" thing. It's just that as a user, I feel that that indication would be useful.
I think we should keep the change where students could see tags of other students' questions. I don't see why that would be necessary.

abcdefguan:
Also, since we're here. How are we going to deal with that privacy issue?

SamChou19815:
> Also, since we're here. How are we going to deal with that privacy issue?

This is a leftover from the past. In the past, we also fetch all questions content but just hide it on student side. Dealing with it is very involved, since it might involve splitting the collection into a full one and a one without question.

Since the firebase change doesn't make things worse, we should probably just proceed.

Shefali-A:
Please don't CC me into these emails.

abcdefguan:
@Shefali-A I think you have to unsubscribe from this thread. I don't know how to remove you from it or why you're involved in it in the first place.
@SamChou19815 I think we might be able to hide this information by using splitting the student and TA functionality (I think the code is too unmaintainable when student and TA functionality are combined) and then using a reduced information version of the FirebaseUser in the student view. However, you can still access this information through the network / firestore.
I think there's nothing we can do about the privacy issue for now, but we could notify Evan about this.

SamChou19815:
> @SamChou19815 I think we might be able to hide this information by using splitting the student and TA functionality (I think the code is too unmaintainable when student and TA functionality are combined) and then using a reduced information version of the FirebaseUser in the student view. However, you can still access this information through the network / firestore.
> 
> I think there's nothing we can do about the privacy issue for now, but we could notify Evan about this.

In this case, hiding no longer makes sense since real hacker can figure that out from network request anyways. Firestore doesn't support projection so there is no easy way to exclude that from requests

abcdefguan:
Makes sense. Though it seems really dumb I can inspect this information through the React components tab lol

PR #316, opened 2020-04-10
Split public and private question data.

rockon999:
### Summary
- Private data is stored in a subcollection/doc at `/questions/{question}/content/private`
- Selectively load question data when isTA is flagged

(more coming soon)

What currently works:
- Student view

What is a WIP:
- TA view

### DIFF
--- a/src/components/includes/AddQuestion.tsx
+++ b/src/components/includes/AddQuestion.tsx
@@ -121,22 +121,25 @@ class AddQuestion extends React.Component<Props, State> {
     public addQuestion = () => {
         if (auth.currentUser != null && this.state.selectedPrimary != null &&
             this.state.selectedSecondary != null) {
-            const newQuestion: Omit<FireQuestion, 'questionId'> = {
+            const batch = firestore.batch();
+            const questionId = firestore.collection('questions').doc().id;
+            const newQuestionSlot: Omit<FireQuestionSlot, 'questionId'> = {
                 askerId: auth.currentUser.uid,
                 sessionId: this.props.session.sessionId,
                 status: 'unresolved',
                 timeEntered: firebase.firestore.Timestamp.now()
             };
-            const newQuestionPrivate: Omit<FireQuestionPrivate, 'questionId'> = {
+            const newQuestion: Omit<FireQuestion, 'questionId'> = {
+                ...newQuestionSlot,
                 answererId: '',
                 content: this.state.question,
                 location: this.state.location,
                 primaryTag: this.state.selectedPrimary.tagId,
                 secondaryTag: this.state.selectedSecondary.tagId
             };
-            firestore.collection('questions').add(newQuestion).then(questionDoc =>
-                questionDoc.collection('content').doc('private').set(newQuestionPrivate)
-            );
+            batch.set(firestore.collection('questionSlots').doc(questionId), newQuestionSlot);
+            batch.set(firestore.collection('questions').doc(questionId), newQuestion);
+            batch.commit();
 
             this.setState({ redirect: true });
         }
@@ -285,6 +288,7 @@ class AddQuestion extends React.Component<Props, State> {
                             + '. Consider adding yourself to a later queue.'}
                         buttons={['Cancel Question', 'Add Anyway']}
                         cancelAction={this.handleXClick}
+                        course={this.props.course}
                         mainAction={() => this.handleJoinClick()}
                         displayShade={this.state.width < this.props.mobileBreakpoint}
                     />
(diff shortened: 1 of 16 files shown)


src/components/includes/SessionQuestionsContainer.tsx
```diff
@@ -38,7 +40,18 @@ const SessionQuestionsContainer = (props: Props) => {
...
 
-    const allQuestions = props.questions;
+    const initQuestions = props.questions;
+    const [allQuestions, setAllQuestions] = React.useState<FireEntry[]>([]);
+
+    React.useEffect(() => {
+        if (props.isTA) {
+            Promise.all(allQuestions.map(q => firestore.collection('questions').doc(q.questionId).get())).then(q => {
+                setAllQuestions(q.map(q => ({ questionId: q.id, ...q.data() } as unknown as FireQuestion)));
+            });
```
SamChou19815, line 50: This will be a terrible `O(n)` number of IOs. We should fetch from questions collection directly instead.

src/components/types/fireData.d.ts
```diff
@@ -70,6 +70,16 @@ interface FireUser {
...
+interface FireSlot {
+    askerId: string;
+    sessionId: string;
+    status: 'assigned' | 'resolved' | 'retracted' | 'unresolved' | 'no-show';
+    timeAddressed?: FireTimestamp;
+    timeEntered: FireTimestamp;
+    questionId: string;
+    questionRef: string;
+}
+
```
SamChou19815, line 82: We should rename it as `FireQuestionSlot` and the collection as `questionSlots` for clarity.

src/components/includes/SessionQuestion.tsx
```diff
@@ -119,42 +134,62 @@ class SessionQuestion extends React.Component<Props, State> {
...
+        firestore.doc(`questions/${this.props.entry.questionId}`).update({
             status: 'resolved',
             timeAddressed: firebase.firestore.Timestamp.now()
         });
+
+        const slot = firestore.collection('slots').doc(this.props.entry.questionId);
+
+        slot.update({
+            status: 'resolved'
+        });
```
SamChou19815, line 171: Use `firestore.batch` to batch the updates

src/components/includes/CalendarView.tsx
```diff
@@ -27,10 +27,21 @@ export default ({ session, sessionCallback, course, user }: Props) => {
...
+        const end = session.endTime.toDate();
+        // TODO This doesn't handle timezones quite right but 
+        // works better than the other somehow...
+        return (selectedDate.getUTCFullYear() === start.getUTCFullYear() &&
+            selectedDate.getUTCMonth() === start.getUTCMonth() &&
+            selectedDate.getUTCDate() === start.getUTCDate()) ||
+            (selectedDate.getUTCFullYear() === end.getUTCFullYear() &&
+                selectedDate.getUTCMonth() === end.getUTCMonth() &&
+                selectedDate.getUTCDate() === end.getUTCDate());
+    });
```
SamChou19815, line 44: What's the rationale for the change?

src/components/includes/AddQuestion.tsx
```diff
@@ -132,7 +132,26 @@ class AddQuestion extends React.Component<Props, State> {
...
+                            status,
+                            questionId: question.id,
+                            timeEntered,
+                            questionRef: question.id
+                        };
+
+                        firestore.collection('slots').doc(question.id).set(newSlot).catch(err => console.log(err));
+                    }).catch(err => console.log(err)) ;
+                });
             this.setState({ redirect: true });
```
SamChou19815, line 155: We should pre-generate an ID using `firestore.collection('questions').doc().id`, and use the same id for both.
Then this can be done in a single batch operation.

src/components/includes/SessionQuestionsContainer.tsx
rockon999: Agreed, luckily this is gone. But we have a similar issue in fetching all the private data. In that case, I'm thinking a query is possible.

src/components/includes/CalendarView.tsx
rockon999: See the updated code, it is actually correct. The current filtering logic is completely broken for me in the late night time hours.

src/components/includes/SessionQuestionsContainer.tsx
SamChou19815: @rockon999 I don't think subcollection query can work. The only possible query that can be done on subcollection is collection group query: https://firebase.google.com/docs/firestore/query-data/queries#collection-group-query
, but it doesn't solve our problem.

Issue #340, opened 2020-04-22
too many notifications - Cached questions !== questions upon re-render of SessionView


ngwattcos:
In branch scott/notifs, in SessionView.tsx: a notification is triggered on any component re-render if at least 1 question is in the queue (page is resized, question "assigned to me," or when a new question is added to queue). This is due to the check cachedPrevQuestions === questions returning false. We need to find out why this is the case.

PR #341, opened 2020-04-22
Add a native destop notification when a new question enters the queue.

ngwattcos:
### Summary 
Adds a native desktop notification when a new question enters the queue. Use 'react-push-notification' to create a notification when.

Also solves issue #340 where a notification is pushed whenever the SessionView component re-renders, whether 1) a question is added (good), 2) web page is resized (bad), or session question state changes/is selected (bad). The problem with checking ```cachedPrevQuestions === questions``` is that, on re-renders, 'questions' is re-populated by objects from Firestore with the exact same data. So 'questions' is a different reference.

This solved by comparing the current list of question ids with the cached question ids in ```prevQuestSet``` - we only care if current list of `questions` has any new questionIds that the ```prevQuestSet``` does not have.

#### The following scenarios were tested:
#####A notification should show up when....
* every time a single question is entered into the queue. Only one notification on the TA/prof's view should pop up.
* a new question is entered into the queue, but questions already exist in the queue

#####A notification should not show up when...
* a student creates a question, but removes their question
* From TA/prof user: resize the window to half width. No notifications should pop up.
* From TA/prof user: click 'Assign to me'. No notifications should pop up.

#####This does not crash...
Safari ios.

### Breaking Changes 
You just need to run ```yarn install``` to get react-push-notification.

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/types/react-web-notification/index.d.ts
+++ b/types/react-web-notification/index.d.ts
@@ -1,0 +1,1 @@
+declare module 'react-web-notification';
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "react-ga": "^2.5.3",
     "react-moment": "^0.8.2",
     "react-numeric-input": "^2.2.3",
+    "react-push-notification": "^1.3.0",
     "react-router": "^5.1.2",
     "react-router-dom": "^5.1.2",
     "react-scripts": "^3.4.0",
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -15,6 +15,7 @@ import * as ReactGA from 'react-ga';
 import moment from 'moment';
 import { Loader } from 'semantic-ui-react';
 
+import { Notifications } from 'react-push-notification';
 import { auth, firestore } from '../firebase';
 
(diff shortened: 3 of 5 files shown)


SamChou19815:
@ngwattcos have you tested in on mobile iOS to ensure it didn't crash?

ngwattcos:
@SamChou19815 I just tested in ios Safari and nothing breaks. Push notifications aren't supported in iOS Safari.

types/react-web-notification/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-web-notification';
```
SamChou19815, line 1: This package is written in TS. https://github.com/fabioshub/react-push-notification
Why are you declaring this?

ngwattcos: Yikes, I forgot to delete this. This was for an older package that didn't have an @types already. I stopped using that package and stashed my changes so I'm not sure why this ended up getting pushed...

PR #358, opened 2020-05-08
Implement cloud functions. Implement OH stats.

abcdefguan:
### Summary 

Added cloud functions to track statistics

Todo:
Automatically add TAs to session upon updating a question state
Restore undo functionality

Done:
Increment number of answered questions when question becomes "resolved"
Increment number of total questions when question is added
Increment total response time when question becomes "resolved"
Increment total wait time when question becomes "assigned"
Increment number of waited for question when question becomes "assigned"

Changes:
Added Cloud functions to project
Deployed cloud function that increments totalQuestions when new question is asked => Tested to be working

Added a number of fields to session
(totalQuestions, assignedQuestions, resolvedQuestions, totalWaitTime, totalResolveTime)
totalQuestions => Total number of questions in session (excluding no-show and retracted)
assignedQuestions => Number of questions assigned to TAs in session
resolvedQuestions => Number of questions resolved by TAs in session
totalWaitTime => Summation of waiting times for all questions (Time between question asked and assignment) in seconds
totalResolveTime => Summation of resolution times for all questions (Time between question assigned and resolved) in seconds
Added timeAssigned field in questions for tracking response time
Made timeAssigned and timeAddressed "public" => Reverted because this is not needed for the cloud function approach
Added default values for all of these fields to session.

The following variables are now automatically tracked with a cloud function and will update based on state changes:
totalQuestions
assignedQuestions
resolvedQuestions
totalWaitTime
totalResolveTime

Implemented rudimentary frontend interface to display OH stats.

Typescript is working correctly in functions folder

### Test Plan 

Checked that the database values of the variables above were correct when questions were assigned and removed

### Breaking Changes 

None

Firebase schema change (requires migration plan)
=> No migration necessary. Firebase increment automatically generates field.
My changes requires a change to the documentation.
Other change that could cause problems (Detailed in notes)
Changes to Cloud Functions

### Checklist

- [ ] My PR adds a @ts-ignore


functions/tslint.json
```diff
@@ -0,0 +1,115 @@
+{
```
SamChou19815, line 1: Don't need to include tslint. We already migrated to eslint with TS support

functions/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+        return db.doc(`sessions/${sessionId}`).update({
+            totalQuestions: increment
+        });
+    });
+
+// This map maps a question status to a tuple of
+// 1. The number of questions this counts as (0 if no show or retracted)
+// 2. The number of assigned questions this counts as (1 if assigned / resolved)
+// 3. The number of resolved questions this counts as (1 if resolved only)
+const questionStatusNumbers = new Map<string, number[]>();
```
SamChou19815, line 40: Use tuple type `[number, number, number]` instead

functions/src/index.ts
```diff
@@ -0,0 +1,101 @@
+import * as functions from 'firebase-functions';
+import * as admin from 'firebase-admin';
```
SamChou19815, line 2: To fix the linter error here, add a yarn install step in CI for the functions folder

functions/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
SamChou19815, line 1: Why there is no yarn.lock for this package.json file

SamChou19815 (dismissed):

functions/src/index.ts
abcdefguan: I don't know how to do this. The .github/workflows file is read from the root directory only. Also, yarn cannot specify a package.json from some other directory. Could you please help with this?

abcdefguan:
@SamChou19815 
functions/index.ts doesn't use Typescript because I had trouble importing fireData.d.ts. Do you know how to do this correctly?

abcdefguan:
Sorry, pressed wrong button

firebase.json
```diff
@@ -1,12 +1,22 @@
...
         "source": "**",
         "destination": "/index.html"
       }
     ]
+  },
+  "functions": {
+    "predeploy": [
+      "npm --prefix \"$RESOURCE_DIR\" run lint",
+      "npm --prefix \"$RESOURCE_DIR\" run build"
+    ]
```
SamChou19815, line 20: Replace that with equivalent yarn command.

functions/src/index.ts
SamChou19815: CI config fixed.

SamChou19815:
> functions/index.ts doesn't use Typescript because I had trouble importing fireData.d.ts. Do you know how to do this correctly?

This will be more involved. The most correct way is to use yarn workspace to allow cross-package code sharing, but the setup is tricky since firebase doesn't support yarn workspace well. Meanwhile you have two options:

- Keep it as it is
- Copy over type definitions that you will use.

ngwattcos (approved):
Thank you for these changes. This is a substantial update for Queue Me In that allows us to implement important features like improved TA promotion and performance tracking, among other features that rely on Cloud Functions. Nice work on the CI/CD.

ngwattcos (commented):
Looks great!

PR #377, opened 2020-09-09
fix(ProfessorRolesTable): Don't automatically default to 'student' ro…

### LINKED ISSUE #376: Professor role drop down doesn't preserve state.
In the professor view when you select the drop down to change a role, if you click out without selecting a role the individual always gets assigned to student.

ngwattcos:
…le when clicking out of role dropdown.

Fixes #376.
### Test Plan 

Open AdminView (/admin) or the professor role table when enrolled as a professor. Click on a TA/professor user's dropdown role, then click outside. User's role should not change.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorRolesTable.tsx
+++ b/src/components/includes/ProfessorRolesTable.tsx
@@ -120,7 +120,11 @@ export default ({ courseId }: { courseId: string }) => {
                         <Table.Cell>{u.lastName}</Table.Cell>
                         <Table.Cell>{u.email}</Table.Cell>
                         <Table.Cell textAlign="right" className="dropdownCell">
-                            <RoleDropdown user={u} course={course} isSelf={u.email === self?.email ? true : undefined} />
+                            <RoleDropdown
+                                user={u}
+                                course={course}
+                                isSelf={u.email === self?.email ? true : undefined}
+                            />
                         </Table.Cell>
                     </Table.Row>
                 ))}
--- a/src/components/includes/ProfessorRolesTable.tsx
+++ b/src/components/includes/ProfessorRolesTable.tsx
@@ -3,10 +3,11 @@ import { Dropdown, Table } from 'semantic-ui-react';
 import * as _ from 'lodash';
 
 import { firestore } from '../../firebase';
-import { useCourse, useCourseUsers } from '../../firehooks';
+import { useCourse, useCourseUsers, useMyUser } from '../../firehooks';
 import { importProfessorsOrTAsFromPrompt, changeRole } from '../../firebasefunctions';
 
-const RoleDropdown = ({ user, course }: {
+const RoleDropdown = ({ ownUser, user, course }: {
+    readonly ownUser?: FireUser;
     readonly user: EnrichedFireUser;
     readonly course: FireCourse;
 }) => {
@@ -19,8 +20,11 @@ const RoleDropdown = ({ user, course }: {
                 { key: 3, text: 'Professor', value: 'professor' },
             ]}
             onChange={(e, newValue) => {
-                if (newValue.value !== user.role) {
-                    changeRole(firestore, user, course, newValue.value as FireCourseRole)
+                const newValueRole = newValue.value as FireCourseRole;
+                if (user.role !== undefined && newValueRole !== user.role) {
+                    if (user.userId !== ownUser?.userId || newValueRole !== "student") {
+                        changeRole(firestore, user, course, newValue.value as FireCourseRole);
+                    }
                 }
             }}
         />
@@ -35,6 +39,8 @@ export default ({ courseId }: { courseId: string }) => {
     const [direction, setDirection] = useState<'descending' | 'ascending'>('ascending');
     const [column, setColumn] = useState<columnT>('email');
     const course = useCourse(courseId);
+
+    const self = useMyUser();
 
     const courseUsers: readonly EnrichedFireUser[] = useCourseUsers(courseId)
         .map(user => ({ ...user, role: user.roles[courseId] || 'student' }));
@@ -110,7 +116,7 @@ export default ({ courseId }: { courseId: string }) => {
                         <Table.Cell>{u.lastName}</Table.Cell>
                         <Table.Cell>{u.email}</Table.Cell>
                         <Table.Cell textAlign="right" className="dropdownCell">
-                            <RoleDropdown user={u} course={course} />
+                            <RoleDropdown ownUser={self} user={u} course={course} />
                         </Table.Cell>
                     </Table.Row>
                 ))}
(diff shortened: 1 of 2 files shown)


src/components/includes/ProfessorRolesTable.tsx
```diff
@@ -18,7 +18,11 @@ const RoleDropdown = ({ user, course }: {
                 { key: 2, text: 'TA', value: 'ta' },
                 { key: 3, text: 'Professor', value: 'professor' },
             ]}
-            onChange={(e, newValue) => changeRole(firestore, user, course, newValue.value as FireCourseRole)}
+            onChange={(e, newValue) => {
```
ewlsh, line 21: I think part of the issue is that we aren't setting an initial `value={user.role}` on this component. This works because the initial value is probably undefined (or similar) - we should still implement this check to avoid unnecessarily updating.

ewlsh (changes requested):

ewlsh:
Another guard we should perhaps also consider is preventing the user from demoting them self. I accidentally did this once, there is no warning and this bug made it a lot worse.

ngwattcos:
@ewlsh I added the guard so that users with the ability to change roles can't demote themself to students.

Interestingly, the issue isn't with user.role, since it can only be "professor" | "student" | "ta". The problem was with newValue.role because it was a string that could be "Student", "TA", "Professor." It turns out if you cast newValue.value as a ```FireCourseRole```, it automatically converts the string to the proper case as well.

src/components/includes/ProfessorRolesTable.tsx
ngwattcos: See below comment.

src/components/includes/ProfessorRolesTable.tsx
```diff
@@ -18,7 +19,14 @@ const RoleDropdown = ({ user, course }: {
...
+                const newValueRole = newValue.value as FireCourseRole;
+                if (user.role !== undefined && newValueRole !== user.role) {
+                    if (user.userId !== ownUser?.userId || newValueRole !== "student") {
```
ewlsh, line 25: I think this line still permits professors to convert themselves to TAs, which isn't ideal.

We could use `deactivate={}` to prevent user interaction and perhaps move the `user.userId !== ownUser?.userId` check out of the component and simply pass it as a boolean.

That will also prevent any interaction and grey out the dropdown.

src/components/includes/ProfessorRolesTable.tsx
ewlsh: I did some digging.

We need to set `defaultValue` instead of `text`. [In the CSS here](https://github.com/cornell-dti/office-hours/blob/e793f6043e46815aa1e8849e5c93ed8e196b7c62/src/styles/professor/ProfessorRoles.scss#L4) it forces the dropdown text to always display in capital case. This is why we get `Ta` instead of `TA`. Part of the confusion of the checks you're doing in `onChange` stems from the fact that we don't set the initial value of the dropdown correctly and we intermix text/value.

We should probably remove that CSS.

src/components/includes/ProfessorRolesTable.tsx
```diff
@@ -106,7 +116,7 @@ export default ({ courseId }: { courseId: string }) => {
                         <Table.Cell>{u.lastName}</Table.Cell>
                         <Table.Cell>{u.email}</Table.Cell>
                         <Table.Cell textAlign="right" className="dropdownCell">
-                            <RoleDropdown user={u} course={course} />
+                            <RoleDropdown ownUser={self} user={u} course={course} />
```
ewlsh, line 119: If we do the userId check here we avoid passing the entire user object.

ewlsh (changes requested):

ewlsh:
@ngwattcos casting can never change the underlying value, what is happening is the way the text field is being displayed is distorted by CSS.

The CSS forces it to always display capital case, which is why we get `Ta` instead of `TA`. We just need to adhere to the dropdown spec better.

src/components/includes/ProfessorRolesTable.tsx
ngwattcos: Agreed, we should probably disable professor level users from editing themselves entirely

ngwattcos:
@ewlsh thanks for the info!

ewlsh:
dropdown info source: https://react.semantic-ui.com/modules/dropdown/

ngwattcos:
@ewlsh Whoops, forgot about this one. I've added the ```defaultText``` value and prevented professors from demoting themselves. I've tested that users won't demote other users by accident when cancelling.

src/components/includes/ProfessorRolesTable.tsx
```diff
@@ -1,26 +1,44 @@
...
+const RoleDropdown = ({ user, course, isSelf }: {
     readonly user: EnrichedFireUser;
     readonly course: FireCourse;
+    readonly isSelf?: boolean;
 }) => {
+    
     return (
         <Dropdown
-            text={user.role}
+            text={toDisplayCase(user.role)}
```
ewlsh, line 25: You should just be able to remove `text=` as the options array associates text with the role value.

src/components/includes/ProfessorRolesTable.tsx
```diff
@@ -1,26 +1,44 @@
...
                 { key: 1, text: 'Student', value: 'student' },
                 { key: 2, text: 'TA', value: 'ta' },
                 { key: 3, text: 'Professor', value: 'professor' },
             ]}
+            defaultValue={user.role}
             onChange={(e, newValue) => {
-                if (newValue.value !== user.role) {
-                    changeRole(firestore, user, course, newValue.value as FireCourseRole)
+                const newValueRole = newValue.value as FireCourseRole;
+                if (isSelf) {
```
ewlsh, line 34: TAs don't need to be able to demote themselves, and they can't promote themselves.

src/components/includes/ProfessorRolesTable.tsx
```diff
@@ -1,26 +1,44 @@
...
+                    if (user.role !== 'professor') {
```
ewlsh, line 35: I would keep this check just-in-case, but I'd also add a `deactivate` property to the dropdown when it is self.

ngwattcos: Gotcha. Now it's ```disabled``` only for your own dropdown.

src/components/includes/ProfessorRolesTable.tsx
ngwattcos: 👍

src/components/includes/ProfessorRolesTable.tsx
ngwattcos: Hopefully the new logic makes this cleaner

triple-jay (approved):
Looks good! I tested on the /admin page with your test plan and it works as it should. Also, your code style looks clear and good.

PR #383, opened 2020-09-11
Fix race condition with onSnapshot.

ewlsh:
### Summary 

The silent onError callback led to unreliable updates of question documents and thus virtual location issues.

From my debugging, we'd sometimes silently fail the onSnapshot call and/or load an un-updated version of the document and per the docs after onError is called onUpdate will never be called again.

Thus, letting the error silently fail proved to be a dangerous foot gun.

I've added a hook which queries for all questions within a given session that the user asked. We then filter this locally. This shouldn't be much more costly than a single document snapshot and it should avoid any potential race conditions.

### Test Plan 

Test locally and in the development environment.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


SamChou19815 (approved):

ngwattcos (approved):
looks good to me!

PR #388, opened 2020-09-15
Fix editing assignment 'active' state.

ewlsh:
### Summary 

Both the name of the assignment _and_ the active state should be updated if they change.

### Test Plan

- Create a new assignment
- Click the edit button
- Change the active/inactive toggle
- Click save
- Ensure the change survives a refresh

### Notes

This was a result of an incomplete implementation when switching to Firestore

### Breaking Changes

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Great!

PR #389, opened 2020-09-18
WIP: Implement assignment deletion support.

ewlsh:
### Summary 

This primarily adds the ability to delete assignments but also incorporates Semantic UI's modal system into the code base instead of relying on browser prompts.

If we don't enable assignment deletion I'd still like to incorporate the modal system.

- Adds a hook for semantic UI's confirmation modal.
- Cleans up open/close CSS
- Removes edit-button toggle

### Test Plan

- Create an assignment
- Delete the assignment
- Verify the deletion survives a refresh and the tag (and its children) are gone from the datastore

### Notes

Seems odd we don't support deleting assignments, *though we should ensure this doesn't break rendering questions with deleted tags* and discuss if deleted tags present a problem.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -20,8 +20,8 @@ import { auth, firestore } from '../firebase';
 
 import AdminView from './pages/AdminView';
 import LoginView from './pages/LoginView';
-import ProfessorView from './pages/ProfessorView';
 import SplitView from './pages/SplitView';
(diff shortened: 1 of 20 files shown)


ngwattcos (approved):
LGTM. Thanks for adding reusable confirmation logic

PR #393, opened 2020-09-21
Fix 'hidden' issues with modality changes in series.

ewlsh:
### Summary 

Hybrid previously technically still required a building location/room number for series.

Essentially, if you switch from virtual to hybrid and don't provide a location it can silently (albeit with a console error) error out.

### Test Plan 

1. Create a new repeating OH (virtual modality)
2. Verify it exists
3. Attempt to switch it to Hybrid
4. Verify the switch happened correctly

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Looks good to me. My questions is what to do with the undefined location value then? Should we prompt the professor to edit the session series to include location?

PR #394, opened 2020-09-21
Only show session checkbox when its a session.

ewlsh:
### Summary 

Previously the session checkbox was shown for every office hour instance.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):

abcdefguan (approved):
Looks good. I mean I helped write this change and I asked you guys to check it.
I don't like the fact that the checkbox is actually doing two different things but is using the same package. I would want to decouple this behaviour, but it's not quite so easy to do so.

PR #396, opened 2020-10-02
Implemented TA comments

abcdefguan:
### Summary 

TA comments have been implemented as per the design provided.

There are a few issues:

- No tooltip with "Comment" displays when mousing over the comment icon. Tooltips are difficult to get code in CSS so I'm leaving it out for now.
- No design was given for how TA comments will look on the student side. It will currently display as "TA Comment: <comment>" at the bottom of the question. This can be improved in future.
- I also used a different gray for the hover of the comment icon because the original gray was not contrasting enough.
- The comment icon is also different from the design icon because semantic UI only provides this comment icon.
- No notification is triggered when a TA issues a comment

New Features:
- Implemented TA comments

### Test Plan 

- Tested that TA can edit a comment, cancel a comment edit and save a comment on the TA side.
- Have not checked student side yet.
- Will probably conduct a Queue Me In test or something in order to check that this works.

### Breaking Changes 

Firebase schema change (requires migration plan)

- [ ] My PR adds a @ts-ignore

No migration plan is necessary because my code handles undefined TA comments

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -3,10 +3,10 @@ import { Icon, Loader, Button } from 'semantic-ui-react';
 import Moment from 'react-moment';
 import * as firebase from 'firebase/app';
 import { useState } from "react";
-import { firestore } from '../../firebase';
-import SelectedTags from './SelectedTags';
 // @ts-ignore (Note that this library does not provide typescript)
 import Linkify from 'linkifyjs/react';
+import { firestore } from '../../firebase';
+import SelectedTags from './SelectedTags';
 
 // TODO_ADD_SERVER_CHECK
 const LOCATION_CHAR_LIMIT = 40;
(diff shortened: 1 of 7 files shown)


src/components/includes/SessionQuestion.tsx
```diff
@@ -323,6 +326,24 @@ class SessionQuestion extends React.Component<Props, State> {
                         </div>
                     </div>
                 }
+                {
+                    this.state.enableEditingTAComment && <div className="TACommentBox">
+                        <div className="commentTopBar">
+                            <img
+                                className="userInformationImg"
+                                src={user.photoUrl || '/placeholder.png'}
```
ngwattcos, line 334: This causes a crash on the student-side. I think replacing with user?.photoUrl should do the trick

src/components/includes/SessionQuestion.tsx
```diff
@@ -267,6 +270,11 @@ class SessionQuestion extends React.Component<Props, State> {
                         <p className={'Question' + studentCSS}>TA Comment: {question.taComment}</p>
                     )}
                 </div>
+                <div className="RightBar">
+                    <button className="commentBtn" onClick={this.toggleTAComment} type="button">
```
ngwattcos, line 274: Have we received designs from the designers for this component yet? If not, I wonder if you could change the implementation slightly to something more like a traditional text box. Here are my suggestions:
- the textbox should already be there so that the TA can readily enter the comments
- The icon is fine; it just bring's the focus to the textbox
- The TA should not have to click "edit" to begin typing into the textbox
- hitting Enter key should automatically save the changes as well

On the other hand, if this is what the designers decided on, then go for it.

ngwattcos (changes requested):

src/components/includes/SessionQuestion.tsx
abcdefguan: This was the design provided. You should talk to the designers about this.

ngwattcos:
Oh no! Something weird is happening with the formatting:
image

ngwattcos (approved):
Thanks for implementing this feature! It looks good to me.

abcdefguan:
The feature has been improved to allow both students and TAs to comment. This PR introduces a huge change and should be tested during subteam meetings to ensure correctness. There are a bunch of minor outstanding CSS issues with this PR which have been highlighted in the Issues.

Issue #400, opened 2020-10-03
Dot Menu is Stretched Inconsistently


ngwattcos:
When Guan merges in #396:

The dot menu is stretched/inconsistent between TA/student-side views. Fixing this requires some design thinking.

TA side
![image](https://user-images.githubusercontent.com/18223523/95001690-55f2a580-059a-11eb-8445-37c7917f1ef7.png)

Student Side
![image](https://user-images.githubusercontent.com/18223523/95001735-e4672700-059a-11eb-929f-96872d986b2d.png)

PR #404, opened 2020-10-08
Wrapped dot menu in a div to fix stretching problem

nada-attia:
### Summary 

Fixed issue #400 by wrapping the dot menu in a div. 

### Test Plan 
Tested this on professor, TA, and student view, and the dot menu no longer appears stretched.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Nice work! I've tested your fix and it works.

PR #405, opened 2020-10-11
Bug fix for inactive classes

triple-jay:
### Summary 
I fixed the error where if a student clicks on an inactive class, it goes to the most recent class.

I just added a conditional statement.

I just added a conditional statement so that it only pushes a new page when the card is not inactive.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Thanks for the simple fix. It works! Nice job.

PR #408, opened 2020-10-18
Jeremytacomments

triple-jay:
### Summary 
Implemented notifications for when TA comments or asker comments

Created new method componentDidUpdate() in component that implements notifications when TA or asker comments.

The method componentDidUpdate() is called whenever the props of the SessionQuestion component change. However, it only fires a notification when the TA comment or user comment is changed, and it only notifies the user and TA, respectively. An imported method addNotification() from react-push-notification is called with the necessary options to create the notification to the user.

### Test Plan 

- Sign into multiple windows as a TA and as a user
- Create a session, and create a question in the queue as a user
- Comment as a TA and see if a notification pops up for the user
- Comment as a user and see if a notification pops up for the TA

### Notes 
This is **not working** yet. As far as I have tested, the blocks are executed whenever a TA or user comments according to some console.log statements , but the problem is, the addNotification() method does not create a notification. This is a problem that persists for me even when I tried to copy their example code on npm in a separate javascript file, which makes me thing either something is wrong with my computer, I am doing something wrong, or the addNotification() method simply does not work.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -5,9 +5,9 @@ import * as firebase from 'firebase/app';
 import { useState } from "react";
 // @ts-ignore (Note that this library does not provide typescript)
 import Linkify from 'linkifyjs/react';
+import addNotification from 'react-push-notification';
 import { firestore } from '../../firebase';
 import SelectedTags from './SelectedTags';
-import addNotification from 'react-push-notification';
 
 // TODO_ADD_SERVER_CHECK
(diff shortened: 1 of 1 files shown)


abcdefguan:
You have some linter issues. It has to do with the number of spaces and tabs.

ngwattcos (changes requested):
Couple things going on here. You should double check that yarn lint passes and that you don't have errors visible in your editor. I have some comments below

ngwattcos (approved):
Looks good!

PR #411, opened 2020-10-19
Nada/notifs

nada-attia:
### Summary 

Implemented a notification modal and banner that prompt the user to enable notifications. These notification messages only appear when the user has not enabled notifications for QMI. 

### Test Plan 

To test, make sure your notifications are turned off, click on a class, and you should see the modal. If you click on an active session, you should see the notification banner. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (commented):
Nice work! I can see both the banner and the modal work when notifications are not enabled. Before I approve, I'm curious if you're able to make this responsively work with mobile pages (when the page is smaller). The modal doesn't show up when the page is resized to mobile size.

EDIT: moving the SVG file to src/media/ should be a quick fix

ngwattcos (approved):
Your changes look great! I've tested on my end as well and the notifications show up for mobil users as well. Nice work.

Issue #407, opened 2020-10-18
Changing Tags to "Inactive" or "Active" Should Update "Child" Tags


ngwattcos:
As in title. Currently, if an assignment (level-1) tag is inactive but changed to active, its secondary (level-2) tags don't update to active. I can imagine the same if vice versa.

How the issue appears on the front-end: you're not able to ask a question in the problem category (because a tag can not be chosen!)
![image](https://user-images.githubusercontent.com/18223523/96357717-73149180-10cd-11eb-9bcb-89ccc04ebdbe.png)

### Suggested Fixes
A hotfix will be tedious because it requires searching for and updating the active/inactive status in level-2 tags that reference the problem level-1 tag as a "parent." The issue is a result of how we store "assignment" (category/level-1 tags) names and "tags" (level-2 tags) in the same collection.

1. Brute force: every update to a tag's status should also modify all tags that reference the to-be-modified tag. This requires searching through all tags for such "children" tags with `.where('parentTag', '==', tagID)`.

2. Another solution is to add a new field tagFamilyID to tags that allows us to filter tags by tag "family" (a tag and all tags parented to this tag will have the same tagFamilyID). While this is more scalable, the actual implementation of this fix is can be quite ugly since data currently exists in the database.

PR #412, opened 2020-10-20
Keep Active Status Children Tags In Sync With Assignment Tag

ngwattcos:
…of parent tags

### Summary 
Fixes https://github.com/cornell-dti/office-hours/issues/407.

The problem: after creating a tag with children tags that are initially active or inactive, editing the active status of the assignment tag later would not change the active status of children tags. This could result in an inconsistent state where the assignment is active but the assignment children tags are not. Thus it would be impossible for a student to submit a question on that assignment.

However, there is still some bad data on the production database. The classes experiencing this issue are recommended to change the active status of a tag, hit save, and change the active status back (and hitting save again).

### Test Plan 
1. For classes where the parent/children tags are in an inconsistent active state: Change the active state and hit save. They should be consistent now.
2. For inactive assignments, change the assignment to active and save. Children tags should be active as well and selectable on the student-side.
3. Vice versa for active assignments.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorTagInfo.tsx
+++ b/src/components/includes/ProfessorTagInfo.tsx
@@ -135,7 +135,6 @@ class ProfessorTagInfo extends React.Component<PropTypes, State> {
                 this.props.childTags.forEach(childTag => {
                     const childTagDoc = firestore.collection('tags').doc(childTag.tagId);
                     batch.update(childTagDoc, {
-                        name: this.state.tag.name, 
                         active: this.state.tag.active
                     });
                 })
--- a/src/components/includes/ProfessorTagInfo.tsx
+++ b/src/components/includes/ProfessorTagInfo.tsx
@@ -132,6 +132,13 @@ class ProfessorTagInfo extends React.Component<PropTypes, State> {
                     name: this.state.tag.name, 
                     active: this.state.tag.active
                 });
+                this.props.childTags.forEach(childTag => {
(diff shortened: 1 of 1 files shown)


src/components/includes/ProfessorTagInfo.tsx
```diff
@@ -132,6 +132,13 @@ class ProfessorTagInfo extends React.Component<PropTypes, State> {
                     name: this.state.tag.name, 
                     active: this.state.tag.active
                 });
+                this.props.childTags.forEach(childTag => {
+                    const childTagDoc = firestore.collection('tags').doc(childTag.tagId);
+                    batch.update(childTagDoc, {
+                        name: this.state.tag.name,
```
triple-jay:
```suggestion
                        name: childTag.name, 
```

triple-jay: This fixes a bug where if you change the status of an assignment from active to inactive, all the tags' names get set to the name of the assignment.

triple-jay (changes requested):
I did test the feature, and it looks good as far as keeping the active status of tags in sync with the assignment. When I set the assignment to active/inactive, the tags did indeed become active/inactive. However, there is a bug in that when the state of the assignment is changed, all the tags' names get changed to the name of the assignment tag. Please look at my comment for your code---I think my small change fixes this bug.

triple-jay (approved):
Looks good now! I see you fixed the bug while I wrote my review :)

PR #413, opened 2020-10-22
Jeremy/newquestionnotif

triple-jay:
### Summary 
I fixed the notification bug that prevented the "Your question is up!" notification from working.

I found that there was a bug in the code that always made the index of myQuestion -1 in the allQuestions array, and I changed it so that it actually returns the actual index of myQuestion. By doing this, I made sure the notification was actually fired when myQuestion has index 0 in the array.

I realized with a few console.log() statements that the useAskerQuestions method was returning objects that had properties that did not exactly match the properties of question objects in the allQuestions array. As a result, myQuestion, which depends on myQuestions which is called from useAskerQuestions(sessionId, myUserId), had different properties compared to question objects in the allQuestions array. This made it so that the original code, allQuestions.indexOf(myQuestion), was always returning -1, which is why the notification was never fired.

Screen Shot 2020-10-21 at 7 51 53 PM

I changed this line to allQuestions.findIndex(elt => elt.questionId === myQuestion.questionId) so that regardless of different properties, it finds the question in allQuestions with the specific id that matches myQuestion. After that, I also changed the notification code to call addNotification() to be consistent with how we are handling notifications with React.

### Test Plan 

- Test the standard case of adding the first question in the queue to see if a notification is fired
- Afterwards, login with two or more student accounts and create two or more questions in a queue---make sure the account you are specifically testing for the notification does not start at index 0 in the queue immediately
- Subsequently delete notifications until the account you are testing with is up, and see if a notification is fired when that question has index 0

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestionsContainer.tsx
+++ b/src/components/includes/SessionQuestionsContainer.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Icon } from 'semantic-ui-react';
 import moment from 'moment';
 import SessionQuestion from './SessionQuestion';
+import addNotification from 'react-push-notification';
 import { useAskerQuestions } from '../../firehooks';
 
 const SHOW_FEEDBACK_QUEUE = 4;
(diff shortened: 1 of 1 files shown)


ngwattcos (approved):
I've tested your code and it works for me! It doesn't send a notification if you're the only person to ask the question, but I think that is okay.

PR #414, opened 2020-10-24
added the netid of students after name in ta view

vladiatrinh:
### Summary 

When there are questions from students in the TA view, the component now also displays the students' netid so the TA can easily contact them if necessary. 

### Test Plan 

I ran QMI locally and created a new account where I was a student. There I added a question and made sure that in my original account, the TA view showed my email address up until the @ symbol. For Cornell emails, this with display their netid properly.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


nada-attia (changes requested):
looks great! but before I can approve, can you fix the linter issue thats causing your build to fail? (run `yarn lint` to see the issue)

edit: also, there seems to be a merge conflict that needs to be resolved

src/components/includes/SessionQuestion.tsx
```diff
@@ -230,7 +230,7 @@ class SessionQuestion extends React.Component<Props, State> {
                                 alt={asker ? `${asker.firstName} ${asker.lastName}` : 'unknown user'}
                             />
                             <span className="Name">
-                                {asker.firstName + ' ' + asker.lastName}
+                                {asker.firstName + ' ' + asker.lastName + ' (' + asker.email.slice(0,asker.email.indexOf('@')) + ')'}
```
abcdefguan, line 233: This will crash if the email somehow doesn't contain @. You could make it more robust if you want, but I think this should be fine.

abcdefguan (commented):
The feature appears to work correctly.
I think we also discussed letting students see TA net IDs when a TA is assigned to them and it would be good to implement that here.
However, I noticed that students can't actually see which TA assigned them. In fact, this actually throws a firebase error where the student doesn't have enough permissions to access the TA's information for some reason. Do you see this bug on your end?

abcdefguan (approved):

nada-attia (approved):

PR #415, opened 2020-10-24
Fix discrepancy between student question number and position of stude…

nada-attia:
…nt in the queue

### Summary 

There was a bug causing the student question number to appear as 0 regardless of the number of students ahead. This bug was caused by trying to find the index of the myQuestion object in the allQuestions list which contains a different type of objects. This always returned -1. I fixed this bug by finding the index of the object in the allQuestions list with the same questionId as myQuestion.

### Test Plan 

Use several accounts to join the queue. The question numbers in each student view should correspond to the question numbers in the professor view (they should no longer be 0). When questions are removed from the queue the numbers in the student view should update accordingly. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestionsContainer.tsx
+++ b/src/components/includes/SessionQuestionsContainer.tsx
@@ -109,7 +109,7 @@ const SessionQuestionsContainer = (props: Props) => {
         return null;
     }, [myQuestions]);
 
-    const myQuestionIndex = allQuestions.find(question => question.questionId === myQuestion?.questionId)   
+    const myQuestionIndex = allQuestions.findIndex(question => question.questionId === myQuestion?.questionId)   
 
     // Only display the top 10 questions on the queue
     const shownQuestions = allQuestions.slice(0, Math.min(allQuestions.length, NUM_QUESTIONS_SHOWN));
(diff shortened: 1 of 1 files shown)


abcdefguan (changes requested):
Does not work.
If you have multiple student questions in the queue and you change the time your question was entered in the Firestore database, you'll find that your order does not change correctly.
If you look at the console you'll find a bunch of errors relating to firebase not having enough permissions to view information. That is the cause of this issue.
The cause of this bug has to do with changes to security rules in the QMI server.
This means that a student cannot see the questions of other students for privacy reasons. (There was a bug related to this and it was patched. This information is not available to students and cannot be seen even if you examine the React component)
As such, allQuestions only contains the student's question and the index is always incorrect.
There's no possible fix to this issue other than some sort of cloud function or a migration to a Node.js backend because a student cannot gain access to information about other students' questions.

nada-attia:
hmm, that's a good point. I see what you're saying, but even if the allQuestions array contained all the questions directly from the database, the old code wouldn't work because it's trying to find the index of an object in an array of different types of objects (see image attached). Even though we can't fully fix the permission restriction issue without migrating the backend, I think maybe we should still fix the indexing issue. 

Screen Shot 2020-10-24 at 11 09 27 PM

abcdefguan:
I don't agree that the allQuestions array contains different types of objects. That array contains only objects of type FireQuestion. In fact if you read the typescript, allQuestions is typed as a FireQuestion[].

abcdefguan:
I found out why this failed on my end. It's caused by data duplication across questions and questionSlots. That was a workaround to fix some question privacy issues we had, but the data duplication is not enforced. Your code should work correctly during normal usage.

abcdefguan (approved):

PR #420, opened 2020-10-25
added message to enable system notifications

ngwattcos:
### Summary 

Added message to enable system notifications.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abcdefguan (approved):
Makes sense to me

PR #431, opened 2020-10-31
Added review session

triple-jay:
### Summary 
Started the feature for adding Zoom review sessions

I created an option for professors to add Zoom review sessions, which has a link to the Zoom review session and allows students to post questions. Professors are asked to input the name as well as the link from the add new office hour button. The Zoom review sessions are stored as a new type of FireSession, namely FireReviewSession, which has modality 'review' and a new property link, which is a string. 

I changed code in a lot of files to make this fix, including creating an ability to create a document in Firestore with this new type of FireSession and rendering this new type when a new review session is created. For now, I just added the review session link as an <a> tag below the cloud function statistics text, as I do not have a specific design for the feature.

When I added FireReviewSession, I got a lot of errors related to the building property being accessed by FireVirtualSession, which is strange, as FireVirtualSession was already an existing type. As a result, I changed a lot of lines of code that related to session.modality to 'building' in session.

### Test Plan 
- Sign in as a professor, and create a new review session from the add office hour button
- Try adding different invalid links to see if it allows you to add invalid links---it should only work when the link starts with "http://" or "https://"
- Once a review session is created, try clicking on the link from the review session to see if it takes you to the right place
- Check on Firestore to see if the new document has a link property and a modality property of 'review'
- See if there are any undesired behaviors related to the new type of session

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionInformationHeader.tsx
+++ b/src/components/includes/SessionInformationHeader.tsx
@@ -31,13 +31,13 @@ const formatAvgTime = (rawTimeSecs: number) => {
     if (isNaN(timeSecs)){
         return "No information available";
     }
-    else if (timeMins === 0){
+    if (timeMins === 0){
         return timeDispSecs + " s"
-    } else if (timeHours === 0){
+    } if (timeHours === 0){
         return timeDispMins + " mins " + timeDispSecs + " s"
-    } else {
-        return timeHours + " h " + timeDispMins + " mins"
-    }
+    } 
+    return timeHours + " h " + timeDispMins + " mins"
+    
 }
 
 const SessionInformationHeader = ({ session, course, callback, user, isDesktop }: Props) => {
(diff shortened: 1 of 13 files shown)


src/components/includes/SessionInformationHeader.tsx
```diff
@@ -87,6 +87,9 @@ const SessionInformationHeader = ({ session, course, callback, user, isDesktop }
                             {' ' + tas.map(ta => ta.firstName + ' ' + ta.lastName).join(', ')}
                         </span>
                     </>)}</p>
+                    {session.modality === "review" ? <div>
+                        <a href={session.link}>Review link</a>
```
ngwattcos, line 91: I would add `target = "_blank" rel = "noopener noreferrer"` to the link.

ngwattcos (changes requested):
Just a small change requested - see the comment.

ngwattcos (approved):

PR #433, opened 2020-11-06
WIP: vladia/csv_upload_buttons

vladiatrinh:
### Summary 

In Professor View
- Removed the import TA and import Professor buttons that used to be inside the table
- Added Manage Roles title and new Import Professors/TAs button outside of table

In Admin View
- Still displays the button in table to maintain same functionality

### Test Plan 

I made sure the buttons in the table are not rendered if user is in Professor View and made sure the buttons are still displayed in Admin View.

### Notes 

The new button does not work yet. In the future, the button should display a new view that allows professors to follow a 3 step process to import profs/TAs into a course.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorRolesTable.tsx
+++ b/src/components/includes/ProfessorRolesTable.tsx
@@ -104,12 +104,12 @@ export default ({ courseId }: { courseId: string }) => {
             </Table.Header>
             <Table.Body>
                 <Table.Row>
-                    <Table.Cell>
+                    {/* <Table.Cell>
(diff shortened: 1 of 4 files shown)


ngwattcos (changes requested):
@vladiatrinh I think we might need to rebase this branch. The reason is that this seems to undo one of the fixes for a bug that automatically demotes professors and TAs to students.

ngwattcos (approved):

PR #434, opened 2020-11-07
Added new discussion button and redesigned modality

triple-jay:
### Summary 
I created a button to separate the add office hour and add discussion buttons. I also added a small redesign to the calendar view to show the type of the session i.e. whether it is an office hour or a discussion.

I created a new button, "Add new discussion" alongside "Add new office hour." This button allows you to create a new review session with a link. Now, the office hours and discussions are more distinguishable, as you can not create a discussion with the add office hour button, and vice versa. I also added a small redesign on the calendar view to show whether a given session is an office hour or a discussion. Finally, I added some small redesign touches to the discussion view by changing how the Zoom review link looks. I did not change too much for the discussion view design, as I know the redesign is probably going to be changed very soon.

New features:
- Added new discussion button
- Separated office hours and discussions into distinct categories
- Created small view on calendar view to show whether session is office hour or discussion
Other changes:
- Changed the location text on the calendar card to be more consistent across modalities
- Small change for review link (CSS and video icon) to make it more visually appealing

### Test Plan 
- Try creating a new discussion with the add discussion button, and see if the link works as expected
- Try editing a discussion
- Try creating a new office hour to see if the existing functionality is maintained
- Try adding tags to see if it still works as expected, as adding tags is interrelated to adding office hours due to reusing the same view for both
- See if the new indicator that shows whether a session is an office hour or discussion looks good

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AddQuestion.tsx
+++ b/src/components/includes/AddQuestion.tsx
@@ -91,7 +91,7 @@ class AddQuestion extends React.Component<Props, State> {
             } else {
                 this.setState({ selectedSecondary: tag });
             }
-        } else if (this.props.session.modality === 'virtual') {
+        } else if (!('building' in this.props.session)) {
             this.setState({ stage: 40 , selectedSecondary: tag });
         } else {
             this.setState({ stage: 30, selectedSecondary: tag });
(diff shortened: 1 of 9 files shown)


ngwattcos (changes requested):
Very nice work! I've tested editing zoom review sessions and it works beautifully on the professor side. Also, I appreciate adding the tag whether it's an OH or discussion.

However, I am not able to join a session as a student, because the text box is unable to be edited. Was this intentional? (I understand that designs and workflow for zoom sessions may or may not be finalized. One thing for sure, we only briefly discussed whether questions in this queue should be public or not.)

image

ngwattcos (approved):
Excellent! Joining, asking (as student), and resolving questions (as TA) works now. Thanks!

PR #435, opened 2020-11-08
Add Unit tests

abcdefguan:
### Summary 

Update Firebase to firebase 8 (This broke some of our code). This is necessary to run firebase emulator.
Add yarn test => Runs unit tests using mocha, chai and enzyme. Unit tests ran in firebase emulator.
This requires a different tsconfig.json in order to play nice with Node.
Setup environment for yarn test and CI/CD
Wrote some extremely simple sample tests

### Notes 

Warning: Eslintrc was changed

import/no-extraneous-dependencies is set to allow devDependencies because we're using certain testing libraries that we don't want to include in the runtime :O

@typescript-eslint/no-unused-expressions is turned off because it does not interact correctly with chai.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
eggcellent

PR #436, opened 2020-11-08
Implement join the queue redesign

nada-attia:
### Summary 

Implemented the join the queue redesign. In this new design, the user no longer has to click on a button to be prompted with information to join the queue; these prompts are always displayed as long as the student hasn't already joined the queue. All the tags are displayed at all times, and the users can switch between assignments without having to unselect them first. The question text box is always visible, but it can only be edited once the user has completed all the required steps. 

### Test Plan 

- Select an assignment, then select a category. The question text box should become enabled. 
- Select an assignment, select a category, then unselect the category. The question text box should become disabled. 
- Select an assignment, select a category, then unselect the assignment. The question text box should become disabled and when the user reselects the same assignment, no tags should be selected. 
- Select an assignment, select a category, select a different assignment, then select a category for that assignment. The question text box should become disabled once the user selects a different assignment, but once they select a category, it should become enabled again. 
- Add the question to the queue, the correct tags should appear (select and reselect different tags; make sure the final selection is what gets displayed), and the tags should be formatted in the same way as the original design (no breaking changes).

### Notes 
- There are merge conflicts that need to be resolved.
- You should test these new changes with different office hour modalities. Make sure nothing is broken/looks too bad; the Figma file didn't contain redesigns for different modalities, so I didn't make any specific changes for them.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestionsContainer.tsx
+++ b/src/components/includes/SessionQuestionsContainer.tsx
@@ -4,10 +4,13 @@ import moment from 'moment';
 import addNotification from 'react-push-notification';
 import SessionQuestion from './SessionQuestion';
 import { useAskerQuestions } from '../../firehooks';
+import AddQuestion from '../includes/AddQuestion';
+import { Loader } from 'semantic-ui-react';
 
 const SHOW_FEEDBACK_QUEUE = 4;
 // Maximum number of questions to be shown to user
 const NUM_QUESTIONS_SHOWN = 20;
+const MOBILE_BREAKPOINT = 920;
 
 type Props = {
     // Session used to update TAs on question answering
@@ -27,6 +30,7 @@ type Props = {
     readonly haveAnotherQuestion: boolean;
     readonly modality: FireSessionModality;
     readonly user: FireUser;
+    course: FireCourse;
 };
 
 type StudentMyQuestionProps = {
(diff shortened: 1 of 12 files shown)


ngwattcos (commented):
I've tested this PR and everything works. Good work on implementing the new queue designs - looks significantly better and honestly more intuitive. I'm leaving behind some comments for you

src/components/includes/SessionQuestionsContainer.tsx
```diff
@@ -212,10 +211,10 @@ const SessionQuestionsContainer = (props: Props) => {
             }
             {shownQuestions && shownQuestions.length === 0 &&
                 <>
-                    <p className="noQuestionsHeading">
+                    {/* <p className="noQuestionsHeading">
                         {props.isOpen ? 'Queue Currently Empty' :
                             props.isPast ? 'Queue Has Closed' : 'Queue Not Open Yet'}
-                    </p>
+                    </p> */}
                     {!props.isOpen ?
```
ngwattcos, line 218: One thing I would change here, instead of using ternary

`!props.isOpen ? <JSX> : null`, you can do
`!props.isOpen && <JSX>`

ngwattcos (changes requested):
I've tested this PR and everything works. Good work on implementing the new queue designs - looks significantly better and honestly more intuitive. But see the comment I left behind.

ngwattcos (approved):
Thanks for addressing the comments. The rest of your code is intuitive and clean. Looks great!

PR #438, opened 2020-11-10
Handle any Notification errors on iOS.

ewlsh:
### Summary 

I don't like the way this library is set up but better to fix this quickly than leave iOS broken.

### Test Plan 

Ask our iOS testers to test before deploy.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Looks good, but we can only test on iOS after merging.

PR #439, opened 2020-11-14
Added new discussion question type as well as upvoting feature

triple-jay:
### Summary 
I created a new react component for a discussion question, adding features such as upvoting to the question.

I first divided the logic to separate FireQuestion into FireOHQuestion and FireDiscussionQuestion, both of which extend FireQuestion. FireDiscussionQuestion shares some, but not all logic with FireOHQuestion. I then created a new React component for discussion sections, which includes a feature to upvote the question. In Firestore, the questions related to discussion questions store an array of users who upvoted the questions. **All users** should be able to see other people's questions anonymously, as seen in the design. A normal user should be able to see their own question as well as the list of questions, while the TA should be able to see the list of questions.

I have NOT yet added the ability for TA's to be assigned to questions, as the design for that has not been yet finalized.

### Test Plan 
- Try adding a question to a discussion section
- Try upvoting or downvoting the question (should start at default of 1 upvote), and signing in with another account to upvote
- Try deleting the new type of question

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
This is great work, Jeremy! The features can already be used in their current form. I've verified that the voting system works. One thing I noticed is that voting can still occur after a discussion session has ended, but this definitely not a huge issue. (probably not a big deal at all since discussion sessions that have passed are probably just abandoned). Nice work!

PR #441, opened 2020-12-02
CSV Upload

vladiatrinh:
### New Professor CSV Upload View 

Created the initial components for CSV upload in the professor view. This includes a 3 step process to allow professors to import TAs and other professors into a QMI course.

- some logic still needs to be added so that the _next_ button conditionally turns blue when the professor uploads a valid csv file

### Test Plan 

- made sure that all next and previous buttons lead to the appropriate view
- even though functionality is not there yet, the buttons don't crash, but instead but lead back to the manage roles table

### Notes 

- added CSVUploadView.tsx and CSVUploadView.scss

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CSVUploadView.tsx
+++ b/src/components/includes/CSVUploadView.tsx
@@ -7,20 +7,10 @@ type PageInfo = {
 }
 
 const CSVUploadView = (
-    { onCancel } : 
-    { onCancel: () => void }
+    { onReturn }: 
+    { onReturn: () => void }
 ) => {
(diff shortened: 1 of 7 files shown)


triple-jay (approved):
Looks great! The design completely matches the Figma sketch, and your code is really well-organized! Your code was really easy to follow due to the clear way you named your functions. I checked to make sure there are no bugs for professors when going back and forth, and the finish button does lead back to the manage roles view.

PR #442, opened 2020-12-06
Small fixes for discussion session

triple-jay:
### Summary 
Added a few small changes to the discussion session

I made the cursor become a clicker hand whenever it hovers over the upvote button, and I fixed a bug that allowed users to upvote questions after a session is expired.

### Test Plan 

- Make sure expired session questions can not be upvoted
- Check to see if hand becomes a cursor hand when it hovers over upvote

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


nada-attia (approved):
Looks good! I tested your changes and I can no longer upvote questions in expired sessions.

PR #444, opened 2020-12-08
Support adding roles to pending users.

ngwattcos:
### Summary 

This PR adds the second part of the ability to assign roles to users who do not exist in the database yet, which vastly simplifies the process of onboarding TAs and other professors for a course.

This PR adds the firebase function onUserCreate, which - upon the creation of a new Queue Me In user - checks the `pendingUsers` collection for a pending user with a matching email. If a `pendingUser` is found, the roles of the pendingUser document are copied over to the newly-created user, and each of the classes specified in the new roles will gain the newly-created user as that specified role. In addition, now that the user has been created, roles can be added to that user normally, so the user's entry in pendingUsers is deleted.

### Test Plan 

1. Create a new fake gmail account.
2. Add a new document in the pendingUsers collection, where the field `email` matches the full email address of the new fake gmail you created and the `roles` field is a map where each entry is a key equal to the document `id` of a course in QMI(-test) and the value is the role, either `ta` or `professor`. Keep this tab open.
3. On a new instance of Queue Me In in incognito mode, log into your new fake gmail account.
4. After a second (hopefully much, much shorter after the cloud function spools up after repeated users), you will gain the roles specified in pendingUsers.
5. Check the entry for your account in pendingUsers. It should have been deleted automatically.

Bug:
This can POTENTIALLY (although highly unlikely) break the 'new user page experience' for a user if they are assigned a `ta` or `professor` role to ALL of the available classes this semester. `Since only editing the course selection exits the user from the 'non-normal'-mode,` combined with the fact that none of the classes are selectable (since the user is a `ta` or `professor` of every available class, `some users will be stuck on the edit page (where isNormalEditMode = false) forever.`

The fix is as follows: if isNormalEdiintMode is false, if the user selects any of the courses (already checked), OR if the user is a `ta` or `professor` of any of the courses, the user should be able to save & exit editing mode.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -7,6 +7,95 @@ admin.initializeApp();
 
 const db = admin.firestore();
 
+/** Adds new roles to a user without them being in QMI's system
+ * Not inclusive: Still need to consider users that are
+ * already in the system. (THIS CASE IS HANDLED BY NOT INCLDUING THEM IN THE
+ * pendingUsers COLLECTIONIN THE FIRST PLACE)
+ */
+exports.onUserCreate = functions.firestore
+    .document('users/{userId}')
+    .onCreate(async (snap, context) => {
+        const userId = context.params.userId;
+
+        // get the user doc
+        const userRef = db.collection('users').doc(userId);
+        const userDoc = await userRef.get();
+        const user = userDoc.data() as FireUser;
+
+        const currentRoles = user.roles;
+        const email = user.email;
+
+        // match this email with a user in the pendingUsers collection
+        const pendingUsersSnap = await db.collection('pendingUsers').where('email', '==', email).get();
+
+        pendingUsersSnap.forEach(async doc => {
+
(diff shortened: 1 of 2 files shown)


vladiatrinh (approved):
Everything seems to work as expected following your testing plan. 
After adding a new document to pendingUsers in QMI-Test Cloud Firestore as TA for info4998, I was able to sign into QMI using my newly created email address, which was correctly identified as role I stated in the document. After signing in, this document disappeared from pendingUsers since I am then no longer a pending user.
I did notice the same edge case you mentioned, where a user may be assigned as a TA for all QMI classes. In this case, the user would be stuck on the Add Courses page. This may be a little confusing for users, since when they first sign in, they have to enroll in a course that they are actually not a TA of first. 
Other than that, everything works as expected from my testing and your code is nicely commented and organized. Looks good to me!

PR #449, opened 2020-12-16
Convert simple components to functional style.

ewlsh:
### Summary 

This was just low hanging fruit from the `a11y` work, this pleases our linter and makes these components simpler.

### Test Plan 

Verify chart components continue to work as expected.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Excellent! This makes me warm inside. Also, component works properly.

PR #450, opened 2020-12-16
Cleanup professor sidebar implementation.

ewlsh:
### Summary 

Don't maintain a "selected" array, let pages pass their type. The "selected" array was unnecessary and made state more complicated for no reason (and relied on indices, which isn't the most readable)

### Test Plan 

Verify all tabs of the professor sidebar continue to work.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Looks good. Everything works as intended.

PR #451, opened 2020-12-16
Fix tag creation UI bugs.

ewlsh:
### Summary 

Tag creation behaves oddly because list insertion and deletion doesn't work correctly when there isn't a unique ID for each element. Try adding and then deleting tags several times in the current UI and you can see the buggy-ness.

This utilizes a simple unique key generation technique to ensure tags function as expected.

### Test Plan 

Add and remove multiple tags without and with saving in the Add Assignment view and verify these bugs are fixed.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Added and removed tags with and without saving without problems. Looks good!

PR #454, opened 2020-12-22
update constants

ngwattcos:
### Summary 
Updates constants to `SP21`, which is parsed as `SP` and `21`. Necessary to test courses now that the semester has ended.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


nada-attia (approved):
Looks good. I tested it and classes from FA20 aren't visible

PR #456, opened 2020-12-22
Update semester checking logic to only check if course semester...

ngwattcos:
Update semester checking logic to only check if course semester matches `constants.CURRENT_SEMESTER` rather than using start/end dates.

### Test Plan 

This unbreaks QMI for winter/inter-semester development

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abcdefguan:
Use moment-tz to spawn the semester startDate and endDate so that they're in ET. By default, I'm not sure if it's system timezone or GMT. Otherwise you can always make sure to leave a nice long gap between the actual semester start/end dates and your start/end dates. Ensure that the semester field has a correct value for all courses in production and test databases before merging (ie do your migration).

abcdefguan (approved):

PR #463, opened 2021-01-19
Revert "CSV Upload"

ngwattcos:
Reverts cornell-dti/office-hours#441

This allows importing TA/professor upload to work while a complete fix is in the works.


abcdefguan (approved):
Makes sense

PR #465, opened 2021-01-22
Scott/fix cloud functions

ngwattcos:
### Summary 
Fixes incorrect references to course documents,. which were previously `undefined`.

### Test Plan 
Same as #444 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abcdefguan (approved):
Looks good to me

PR #469, opened 2021-01-23
Fix Daylight Saving Time Issues

abcdefguan:
### Summary 

Replace ad-hoc time computations with moment
Force office hours to be created in America/New_York timezone instead of system timezone
Fix issues around OH times shifted due to daylight saving time
Changes to code adding new OH and updating existing OH times

### Test Plan 

Test cases for helper function that generates the date range can be found in time.test.tsx
OH session generation code was tested manually.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "lodash": "^4.17.15",
     "mocha": "^8.2.1",
     "moment": "^2.22.2",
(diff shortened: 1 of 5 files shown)


ngwattcos (approved):
LGTM!

PR #473, opened 2021-01-25
Implement TA/professor CSV upload and manual input

nada-attia:
### Summary 

Implemented new CSV upload and manual input. Professors can now input TAs and other professors by uploading a CSV file which gets parsed and passed to the `importProfessorsOrTAsFromCSV` function. They also have the option to manually input TA/professors by providing their emails and roles. If the user exists in the QMI database they will get added to the course/their role in the course will be updated accordingly otherwise they will be marked as a missing user and will get added to the `pendingUser` collection. 

This PR addresses issue #467. 

### Notes 

While implementing this feature, I realized that the error messages in the designs don't comprehensively cover the different cases so I added more messages. Also according to the example table in the design (and this PR), the roles that can be added are professor, TA, head TA, consultant, and student, but the only roles that are currently accepted are "Professor" and "TA." I kept the example table in this PR consistent with the designs, but I think it should be updated to reflect that only "TA" and "Professor" are acceptable inputs. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Thank you for fixing the edge cases! It runs beautifully now and integrates well with our firebase cloud function. Looks good to me!

PR #475, opened 2021-01-25
Fix edge case: newly-promoted pending TA/prof can't leave onboarding page.

### LINKED ISSUE #464: !!! Front-end Issue Preventing Users from Saving Enrolled Courses on Home Page
If a new user logs in and they are promoted to a ta/professor of a course, and only that course is available to sign up (on the home/edit pages), then the new user can't "save" their changes and move ahead to the main page.

This issue I discovered from PR #444:

> This can POTENTIALLY (although highly unlikely) break the 'new user page experience' for a user if they are assigned a ta or professor role to ALL of the available classes this semester. Since only editing the course selection exits the user from the 'non-normal'-mode, combined with the fact that none of the classes are selectable (since the user is a ta or professor of every available class, some users will be stuck on the edit page (where isNormalEditMode = false) forever.
> 
> The fix is as follows: if isNormalEdiintMode is false, if the user selects any of the courses (already checked), OR if the user is a ta or professor of any of the courses, the user should be able to save & exit editing mode.

ngwattcos:
### Summary 
Fixes #464 by also tracking `numCoursesWithRoles`, which is the number of courses where the user is either a `professor` or a `ta`.

When not in normal editing mode (`!isNormalEditingMode` is when user is onboarding classes for the first time), we now use `numCoursesWithRoles` + `coursesToEnroll` + `coursesToUnenroll` === 0 as the check for whether user is able to "enroll"

### Test Plan 

Add an entry to pendingUsers, with a new email with all role assigned to all available classes this semester. Then log into QMI using this Google account (for the first time) and wait to get promoted to ALL available classes. The enroll button should still be available.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CourseSelection.tsx
+++ b/src/components/includes/CourseSelection.tsx
@@ -130,7 +130,7 @@ function CourseSelection({ user, isEdit, allCourses }: Props): React.ReactElemen
         history.push('/home');
     };
 
-    const selectedCoursesString = (selectedCourses.length === 0
+    const selectedCoursesString = (selectedCourses.length + numCoursesWithRoles === 0
         ? 'No Classes Chosen'
         : selectedCourses.map(c => c.code).join(', '));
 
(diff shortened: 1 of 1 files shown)


nada-attia (approved):
i tested it and it looks good to me

PR #494, opened 2021-02-27
Prompt user to confirm/cancel before refreshing/closing window

nada-attia:
### Summary 

When a user tries to refresh or close the QMI, they will be prompted to confirm/cancel their action. 

(also fixed some minor css issues)

### Test Plan 

Try refreshing or closing the tab.
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Looks good to me! Thanks for the quick fix. Also thanks for removing some old commented out code, man it should not have been there!

Issue #491, opened 2021-02-25
Convert <AddQuestion> component to a functional style.


ngwattcos:
While this may have been acceptable a few years ago, this must be converted to functional hooks.

![image](https://user-images.githubusercontent.com/18223523/109106988-f179e380-76fe-11eb-95b7-985f97b991f0.png)

PR #500, opened 2021-03-05
Convert AddQuestion component to functional React hooks

nada-attia:
### Summary 

Addresses issue #491. Converted AddQuestion from a class component to a functional component (to keep codebase more consistent). Removed class, constructor, public methods, etc. and replaced them with React hooks equivalent. 

### Test Plan 

Try adding a question by selecting different tags. There should be no breaking changes.
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AddQuestion.tsx
+++ b/src/components/includes/AddQuestion.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { useState, useEffect } from 'react';
 
 import { Redirect } from 'react-router';
 import { Icon } from 'semantic-ui-react';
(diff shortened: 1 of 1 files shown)


ngwattcos (commented):
Excellent work, Nada! I'm holding off on approving this for the moment because we don't want to accidentally revert your last PR, which fixed the location issue for hybrid sessions.

nada-attia:
good catch - i forgot to pull before working on this pr. i fixed the location issue.

ngwattcos (approved):
Perfect! Now this PR keeps the new fix for missing hybrid location/URL. Thanks Nada!

PR #501, opened 2021-03-06
Updated import for logo in TopBar and fixed styles for new aspect ratio

abagel21:
### Summary 

Logo Updated

Updated the import statement within the TopBar.tsx file, and then changed the width styling within TopBar.scss because the aspect ratio of the previous navbar logo was 2:1 whereas this logo is 1:1. It is worth noting that this logo is thus slightly taller than the previous (~4px taller) and takes slightly more vertical space. This appears to have very little affect on the spacing of the other elements in the navbar, however.

### Test Plan 

Changed window size, both width and height, and observed that expected behavior occurred with resizing (and TopBar disappearing at certain width). Navigated throughout the website to make sure there is no unexpected behavior with the logo.

### Checklist

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/TopBar.tsx
+++ b/src/components/includes/TopBar.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { Icon } from 'semantic-ui-react';
 import { logOut } from '../../firebasefunctions';
-import Logo from '../../media/Logo.svg';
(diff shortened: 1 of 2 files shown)


ngwattcos (approved):
Excellent work, Alex! I've tested your code and it's good to see the familiar redesigned logo back

PR #503, opened 2021-03-06
Add missing transitions in add question state machine 

nada-attia:
### Summary 

There were a few missing transitions in the add question state machine that were causing the following (and similar) bug: 
- select a primary tag 
- select a secondary tag 
- reselect a primary tag (which will clear the secondary tag)
- you can still proceed to enter a location and question and the join button seems clickable

The state machine is outlined [here](https://github.com/cornell-dti/office-hours/wiki/AddQuestion-Component#state-machine-outline) 

### Test Plan 

When you unselect a primary tag, secondary tag, clear location, or question, the join the queue card goes back to the step that you changed (everything else gets cleared)
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AddQuestion.tsx
+++ b/src/components/includes/AddQuestion.tsx
@@ -69,11 +69,14 @@ const AddQuestion = (
 
     const handlePrimarySelected = (tag: FireTag | undefined): void => {
         if (selectedPrimary) {
+            setLocation('');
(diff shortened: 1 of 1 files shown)


ngwattcos (approved):
Looks great to me! Thanks for solving a problem that we didn't even know we had.

PR #506, opened 2021-03-09
Fix lint issues causing deploy check to fail

nada-attia:
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
LGTM!

Issue #499, opened 2021-03-05
Stretched header when location/URL is long


ngwattcos:
Fix: Do not display if link. Instead, display message "see Zoom link."

![image](https://user-images.githubusercontent.com/18223523/110049963-df143100-7d20-11eb-9d0f-a9afae37b72e.png)

PR #508, opened 2021-03-13
Add material-ui package and fix session header using grid system

nada-attia:
### Summary 

This pr fixes issues such as #499. The addition of the material ui grid system fixes the session header's appearance by allowing the size of header items to change as the screen size changes. 

Note: When checking out this branch, run `yarn` to install the new material-ui dependency.

### Test Plan 

This pr does not add any new functionality. 
When testing, make sure that the header items look good when you change your screen size. Try adding a long title or location when creating a new office hour session, the text should get cut off and there should be no stretching. 
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "proxy": "http://localhost:3001",
   "dependencies": {
+    "@material-ui/core": "^4.11.3",
     "@nivo/bar": "^0.59.2",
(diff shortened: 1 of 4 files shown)


ngwattcos (approved):
Great work, Nada! I think this is the most sensible solution as the link should still be available.

PR #510, opened 2021-03-13
Added loading for new users during cloud function latency

abagel21:
### Summary 

Added a new observable hook for a pendingUser document associated with the current user, so that when a user first logs in, it verifies whether the cloud function will run.

If the user has no associated pendingUser document, the course selection view loads as normal. If the user has an associated pendingUser document, the <Loading /> sematic ui React component notifies the user that there is logic running such that they cannot access the page. 

### Test Plan 

Tested loading with a new user both with and without a pendingUser document associated.  Also verified that the other course selection (non-editing view) is unaffected. 

### Notes 

Possible future development may involve a new component for this specific case that is less ambiguous. The loading screen currently is just the loading circle, while perhaps a custom message may alleviate new user concerns.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/pages/CourseEditView.tsx
+++ b/src/components/pages/CourseEditView.tsx
@@ -1,12 +1,14 @@
 import * as React from 'react';
 import { Loader } from 'semantic-ui-react';
 import CourseSelection from '../includes/CourseSelection';
-import { useMyUser, useAllCourses } from '../../firehooks';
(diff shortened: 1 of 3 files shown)


ngwattcos (approved):
Great work, Alex! We can probably iterate on this in the future.

PR #518, opened 2021-03-20
Move addQuestion backend logic to firebase functions file

nada-attia:
### Summary 

This PR moves the backend logic for adding a question to the queue from the AddQuestion component to the firebase functions file.  

### Test Plan 
This PR does not introduce any new code. Try adding a question -- the functionality should be the same.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Works the same as it has before!

PR #519, opened 2021-03-20
Refactor SessionQuestion backend logic

nada-attia:
### Summary 

Removed the backend logic responsible for retracting a question, marking a question as resolved/no show/dont know, and commenting on a question from the SessionQuestion component to the firebasefunctions file.

### Test Plan 

- Remove a question you added 
- Mark a question you added as done (from TA side)
- Mark a question you added as no show (from TA side) 
- Comment on a question from both the TA and student side 

Additionally, you can check the question documents in firestore. The status field for each question doc should be updated accordingly.  

There should be no breaking changes. 
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -1,13 +1,21 @@
 import * as React from 'react';
 import { Icon, Loader, Button } from 'semantic-ui-react';
 import Moment from 'react-moment';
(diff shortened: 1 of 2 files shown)


ngwattcos (approved):
Beautiful! It works with the new state machine logic. Nice work!

PR #520, opened 2021-03-20
Made minor changes to landing page

triple-jay:
### Summary 

Made very minor fixes to the landing page.

I added the second curved ribbon right above the "for professors" header, I replaced the demo image with a new one, and I also added the logo in the footer. 
UPDATE: I forgot to add that I also removed the animation that fades in the login button and queue me in headers. This was bugged for some reason in the production version, so I got rid of it.

### Test Plan 
- Check that the ribbon, new demo image, and logo all look good for both desktop and mobile devices
- UPDATE: Check that the fade in animation for the login button and queue me in headers no longer works

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/components/pages/LoginView.tsx
```diff
@@ -86,6 +87,7 @@ const LoginView: React.FC = () => {
                         </div>
                     </div>
                 </div>
+                <img src={SecondWhiteStrip} className="secondRibbon" />
```
nada-attia, line 90: can you add an alt tag here? 
missing alt tags cause cd checks to fail

ngwattcos (dismissed):
Looks great! Thanks for presenting! Just be sure to fix the mere conflicts.

abcdefguan (approved):
Landing page looks good and is responsive for large screens. Great work. I like the hover effect on the QMI Dashboard and the hover mouseover effect on some of the pills.

ngwattcos (approved):
Re-reviewing to approve your new changes. Looks good, and thank you for adding alt tags to images! For some reason these are only warnings locally but count as errors in the CD/CD pipeline.

PR #522, opened 2021-03-20
Fixing issue #497, refactoring AdminView into multiple components, and adding some styles to AdminView

### LINKED ISSUE #497: Improve Admin View Interface
While the admin view is completely functional, it needs some front-end rework to prevent components from overlapping (with the "Archived Courses" list) if the list of classes is large.

Good task for a future TPM.

image

abagel21:
### Summary 

I utilized the Material UI package to refactor the grid system for the courses and make it more responsive, styled some of the buttons, made the courses into Material UI Cards, refactored out the multiple components in the AdminView file, and added the TopBar to the admin page.

- Replaced the current grid style system with Grid components from @material-ui, which also makes the grid responsive--four courses per row on a very large screen, three per large, two per small, one per extra small. This fixes #497.
- Refactored the code within AdminView, which contained four functional components internally. These have been separated into /includes/[AdminCourseCard, AdminCourseCreator, AdminEditableCourseCard, AdminReadOnlyCourseCard].
- Updates some styles for the AdminView page, such as turning the course cards into <Card> components from @material-ui, making the button styles more consistent with the rest of the site, and shifting the import buttons outside of the table of ProfessorRolesTable
- Added the TopBar to the admin page utilizing a new optional boolean prop, admin. If admin is true, it mimics the behavior of a student TopBar with the [edit] still visible in the dropdown, so all of the site can be navigated to from the admin page.

### Test Plan 
These are the steps I followed to test the changes:
- Test that the AdminView Grid displays all information without overlapping on a variety of screen sizes/numbers of classes
- Test that the TopBar correctly displays the right information and navigates to the right pages
- Test that the admin page functionality still works with the refactoring (None of the functionality code has changed)
- Test that the new styles do not interfere with functionality and are responsive along with the Grid

### Notes 

There are a couple of other small style tweaks possibly worth implementing, such as the multiline dropdown at certain screen sizes in the roles table, or the structure of data in the cards. Also, functionality in editing for the "Save" and "To Read Only" should likely be combined, but because the admin page is internal perhaps it is unnecessary.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AdminCourseCard.tsx
+++ b/src/components/includes/AdminCourseCard.tsx
@@ -1,0 +1,27 @@
+import React, {useState} from "react";
+import AdminReadOnlyCourseCard from "./AdminReadOnlyCourseCard";
+import AdminEditableCourseCard from "./AdminEditableCourseCard";
+import ProfessorRolesTable from '../includes/ProfessorRolesTable';
+import {Card} from "@material-ui/core";
+
+const AdminCourseCard = ({ course }: { readonly course: FireCourse }) => {
+  const [isEditingMode, setIsEditingMode] = useState(false);
+  const [showRolesTable, setShowRolesTable] = useState(false);
+  return (
+      <Card className="course">
+          {!isEditingMode && <AdminReadOnlyCourseCard course={course} />}
+          {isEditingMode && <AdminEditableCourseCard course={course} />}
+          <div>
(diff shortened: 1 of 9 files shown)


ngwattcos (approved):
Works beautifully! Thanks for making the admin view a joy to use :')

PR #525, opened 2021-03-27
fixed series start times when start has passed

rmz38:
### Summary 

Fixes issue where first session of a series of office hours is not added but succeeding ones are.

Fixed by making a copy of the start time with end time (as moment) added to check if end time has passed  in the firebase functions file.

### Test Plan 

Tested using various start times that are before and after the current time

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore
^ not necessary as it is a small change code wise


src/firebasefunctions.ts
```diff
@@ -75,6 +75,7 @@ export const createSeries = async (
 
     const startTime = moment(sessionSeries.startTime.toDate()).tz("America/New_York");
     const endTime = moment(sessionSeries.endTime.toDate()).tz("America/New_York");
+    console.log(endTime);
```
abcdefguan, line 78: Remove the console print

src/firebasefunctions.ts
```diff
@@ -88,7 +89,8 @@ export const createSeries = async (
 
     datesToAdd.forEach( (sessionStart) => {
         // Do not add sessions before today or course start
-        if (sessionStart.isBefore(now) || sessionStart.isBefore(courseStartTime)){
+        const checkStart = sessionStart.clone().add(endTime.diff(startTime)).isBefore(now);
```
abcdefguan, line 92: Use moment(sessionStart) to clone the moment. Add duration instead of endTime.diff(startTime). So moment(sessionStart).add(duration).isBefore(now).

abcdefguan (changes requested):
There are a bunch of tests testing the date generation algorithm in time.test.tsx. I haven't pulled the entire date generation into that test, but you could pull the date generation state into that, then update the test to test your new behavior.

abcdefguan (approved):

PR #527, opened 2021-04-03
Refactor backend logic from SplitView and SessionView

nada-attia:
### Summary 

Moved the logic in SessionView that is responsible for updating the virtual location of a question after a TA assigns themself to that question to the sessionQuestion file. 

Moved the logic in SplitView that is responsible for removing a student's question from the queue to the sessionQuestion file. 

### Test Plan 

Ask a question then remove it, the question should be marked as retracted in the database (verify by looking at data in firestore) 

As a TA, assign a student's question to yourself. The student should be able to navigate to your virtual location when they click on join. 
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionView.tsx
+++ b/src/components/includes/SessionView.tsx
@@ -7,12 +7,13 @@ import SessionQuestionsContainer from './SessionQuestionsContainer';
 
 import { useCourseTags, useCourseUsersMap, useSessionQuestions, useSessionProfile, 
     useAskerQuestions } from '../../firehooks';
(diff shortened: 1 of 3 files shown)


triple-jay (approved):
Great work! I tested that if you add a question and remove it, the status of the question becomes retracted. I also tested that the student can join a TA's zoom link once the TA is assigned to the question.

I really am impressed with how the files look a lot more organized now thanks to your work! Makes everything look a lot cleaner.

PR #532, opened 2021-04-04
Added Blog Post CRUD firebasefunctions and front end for creation, editing, and deleting

abagel21:
### Summary 

Created CMS for making, editing, and deleting blog posts.

- Created blogPost type
- Added a security rule for blogPosts (both read and write require isAdmin, will refactor with admin permissions)
- Created firebasefunctions for creating, editing, and deleting blogPosts in the back end
- Created fire hook for reading all the blogPosts currently in the collection
- Created a front end form for creating blog posts utilizing the firebase function
- Created a front end grid for displaying blog posts that allows editing and deleting

### Test Plan 
These were the steps I took
- Test making a single blog post with normal/expected amounts of information
- Test adding number of blog posts with different numbers of listItems and different size titles/descriptions
- test deleting all blog posts
- test submitting the add blog post form without a field properly filled out (empty title, description, or listItem)

### Notes 

When I have time in the future I will likely make the add blog post form prettier.

### Breaking Changes 

- [ ] My PR adds a @ts-ignore


ngwattcos (dismissed):
Awesome work! I've tested the create, edit, and delete functionality and it just works! I also really appreciated how you don't create posts if some fields are missing, and you ask an "Are you sure?" if you try to delete a post. Clean code as well, nice job! I am leaving a few comments for you however

src/components/includes/AddBlogPost.tsx
```diff
@@ -0,0 +1,84 @@
+import React, {useState} from 'react'
+
+import {addBlogPost} from '../../firebasefunctions/blogPost';
+import { firestore, auth } from '../../firebase';
+
+
+const AddBlogPost = () => {
+    const [title, setTitle] = useState("");
```
ngwattcos, line 8: Be sure to change all instances of double quotes to single quotes (anything that is not JSX template).

ngwattcos (changes requested):
Awesome work! I've tested the create, edit, and delete functionality and it just works! I also really appreciated how you don't create posts if some fields are missing, and you ask an "Are you sure?" if you try to delete a post. Clean code as well, nice job! I am leaving a few comments for you however. Another comment here: what are your thoughts on adding a type that represents a Firestore timestamp? (so this allows us to track date information). On the other hand, I imagine that the title can be used for that purpose.

ngwattcos (approved):
Looks good to me now! Thanks for the fix.

PR #534, opened 2021-04-10
Remove a redundant (and unused) logout menu.

ngwattcos:
### Summary 

If you were to checkout `master` and change the following code in `CalendarHeader.tsx`
`const [showCourses, setShowCourses] = React.useState(false);`
to 
`const [showCourses, setShowCourses] = React.useState(true);`
You can see that there is a redundant, unused, and otherwise disabled menu:
| before | after |
| ----------- | ----------- |
| image  |  image |

### Test Plan 

Just test that logout and feedback buttons work as normally. The removed code was a menu that did not display, so there should be no functionality changes.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CalendarHeader.tsx
+++ b/src/components/includes/CalendarHeader.tsx
@@ -1,7 +1,4 @@
 import * as React from 'react';
-import { Icon } from 'semantic-ui-react';
-
-import { logOut } from '../../firebasefunctions/user';
 import { useMyCourses } from '../../firehooks';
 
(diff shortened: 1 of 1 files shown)


nada-attia (approved):
i tested your changes and the logout menu works well on all the different pages. thanks for removing usused code

abcdefguan (approved):
Seems to be a duplicate menu that was controlled here in the past but is controlled by top bar now. Well, at least we don't need Ryan to figure out what's the purpose of that code anymore...

PR #535, opened 2021-04-10
Remove useEffect hooks that contain backend logic 

nada-attia:
### Summary 

Components that were refactored are:
- ProfessorTagsTable 
- ProfessorDashboardView 
- ProfessorPeopleView 
- ProfessorView

These components are responsible for displaying the professor side.

### Test Plan 

- Test the tags section of the professor view. You should be able to add and remove tags as normal.
- Test the people section of the professor view. You should be able to see data for the office hour session (should not tell you that there are no office hour session available to view)
- Test the dashboard view. You should be able to switch between tags. 
- Test professor view. Office hour sessions should be visible in the calendar view. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorTagsTable.tsx
+++ b/src/components/includes/ProfessorTagsTable.tsx
@@ -1,15 +1,10 @@
 import * as React from 'react';
 import ProfessorTagsRow from './ProfessorTagsRow';
 
-import { firestore } from '../../firebase';
-import { useQuery } from '../../firehooks';
-
-const getQuery = (courseId: string) => firestore
-    .collection('tags')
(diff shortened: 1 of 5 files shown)


ngwattcos (approved):
Works great! Tested all the views and all functionality seems to be intact. But is this the original UI? (it seems to be shifted to the left)

image

nada-attia:
> Works great! Tested all the views and all functionality seems to be intact. But is this the original UI? (it seems to be shifted to the left)
> 
> image

I also noticed that the UI looks different a while ago. I didn't change any code that would affect the UI in this PR.

PR #539, opened 2021-04-10
Add missing dependencies causing cd check to fail in firehooks

nada-attia:
### Summary 

PR #535 introduces warnings that cause cd check to fail... we should probably add a build job to ci so we can catch these warnings before prs are merged into master 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Looks good to me! Thanks for the fix, Nada!

PR #541, opened 2021-04-14
Logout Bar Click Off

pooja654:
### Summary 

 The Logout Bar now disappears when users click off (as opposed to only when users click it again). Changed this in TopBar.

### Test Plan 

 I tested on Firebase with yarn start. I also ran yarn lint to check for errors.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Excellent work, Pooja! I've tested your code and the menu disappears when clicking off the top bar, but all the buttons in the menu still work. Congratulations on your first feature PR!

PR #543, opened 2021-04-17
removed firebase logic from calendar files

rmz38:
### Summary 

Firebase logic is removed from the calendar and course files into their own files under the firebase functions folder.

### Test Plan 

I tested changes by adding and removing courses as well as interacting with the calendar, as the changes were small by moving a few lines of code into other files.


ngwattcos (commented):
Just a comment here, but getQuery in `calendar.ts` is okay to be in the original source file.

ngwattcos (approved):
Great work on this PR, Rose! I've tested the functionality and it seems that nothing unexpected has changed.

Issue #526, opened 2021-04-02
App outright fails to load on iPhone / iPad


abcdefguan:
Reported by a student on iPhone. Reported to be working in Feb, but not working now.
Tested on iPad with iOS 14.4, Chrome and Safari

App displays a blank screen after logging in

ngwattcos:
@pooja654 and I are not able to recreate this bug on iPhone. I remember a TA complaining of this issue a while back. Can we get more information on how the bug occurs?

abcdefguan:
When trying to login, It points me towards that google login link and lets me fill up my details. After that's done, it displays a loading icon for a bit and it gives me a blank screen that's the same color as the loading icon. It failed on 2 separate iPads running iOS 14.1 and 14.4 on both Safari and Chrome on my end.

ngwattcos:
That is quite curious. It is possible that some calls to the notification API have not been caught, although I'm puzzled on why it works on half of iOS devices (I have iOS 14.4, safari). Perhaps it is because I didn't have to log in and the authentication process is bugged.

abcdefguan:
That's odd. Maybe it has to do with the authentication process. (??). See if you could borrow an new iPad / iPhone or clear your cache and cookies and see if the problem appears.

PR #544, opened 2021-04-17
Fix notification reference error for ios devices

nada-attia:
### Summary 

Fixes issue #526. The notification reference error was caused because there is no variable named `Notification` (this variable is stored in window) so it must be referenced using `window?.Notification`

### Test Plan 

Connect ios device to macbook with a cable. Run `ifconfig` in terminal to find the ip address that you need to run localhost on ios device. Open `http://<ip address>:3000 `. Sign in** and select an office hour session. The app should redirect you to that session (and should not crash). 

** You need to authorize the domain that you're using to sign into the app using firebase. To do this, go [here](https://console.firebase.google.com/u/1/project/qmi-test/authentication/providers) and add the ip address to authorized domains.
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/pages/SplitView.tsx
+++ b/src/components/pages/SplitView.tsx
@@ -143,8 +143,9 @@ const SplitView = (props: {
                                 Please select an office hour from the calendar.
                                 <p> </p>
                                 <p> </p>
-
-                                {(Notification !== undefined) && Notification.permission === "granted" && (
+                                {("Notification" in window &&
+                                window?.Notification !== undefined) && 
+                                window?.Notification.permission === "granted" && (
                                     <div className="warningArea">
 
                                         <div>


ngwattcos (approved):
Tested on my iPhone. I get as far as the login page, because after I'm being prompted to login on a new page, I'm not redirected. I'm assuming that this will work because the login window is located on qmi.firebaseapp.com......, which does not have the updated code, so I will approve for now.

nada-attia:
this might be because you didn't authorize the domain (the ip address which corresponds to ur local host) that you're using to log in to qmi. you can authorize it [here](https://console.firebase.google.com/u/1/project/qmi-test/authentication/providers)

PR #545, opened 2021-04-17
Removing firestore from front end components and db param from firebasefunctions for BlogPosts

abagel21:
### Summary 

Refactored backend logic out of recently completed components.
- removed firestore from AddBlogPosts.tsx
- removed firestore from BlogPostInternal.tsx
- removed the db parameter from firebasefunctions in line with refactoring firestore out of the front end components

### Test Plan 

Just ensuring the components still perform CRUD operations as expected would be sufficient, but the change should have zero impact on functionality considering it is just refactoring.

### Notes 

Please let me know if I should also remove the {auth} import within the front end components.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rmz38 (commented):
Everything seems fine to me, if Scott or anyone else wants to approve I'd agree.

ngwattcos (approved):
Looks good to me! Thanks for removing firestore from the components.

PR #548, opened 2021-04-24
Landing page redesign

triple-jay:
### Summary 
Worked on landing page redesign and made small changes to the footer to make it look more centered.

I worked on the new landing page redesign for both desktop and mobile devices. I also made the text in the footer look more centered vertically.

### Test Plan 
- Make sure new redesign looks good on both desktop and mobile devices
- Make sure footer looks centered vertically as I described

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


nada-attia (approved):
the new homepage looks great :)

one small thing: would it be possible to move the text down a little bit, so the image doesn't cover it when it moves down? 

Screen Shot 2021-04-30 at 5 28 52 PM

PR #549, opened 2021-05-01
Move Firebase-related code from some professor files into a file within firebasefunctions

triple-jay:
### Summary 
I moved Firebase-related code from some professor files into a new file session.ts within firebasefunctions.

I moved code that either changed or read data from Firestore in the files ProfessorCalendarRow, ProfessorOHInfoDelete, and ProfessorOHInfo into a new file session.ts. I was also asked to move code that changed or read data from ProfessorDelete and CalendarDateItem, but I did not find anything in particular in those files that did either of those operations. If I missed something related to either of these operations in these files, please let me know!

### Test Plan 

Note that for all of these tests (except the last bullet point since it doesn't really apply), I am talking about a **single** office hour and not repeated office hours.
- Try adding an office hour and see that it is added on Firestore
- Delete an office hour and check that it is deleted on Firestore
- Update an office hour (e.g. by updating the time) and see that the correct fields are updated on Firestore for that office hour
- Check to see that in the professor dashboard in the manage hours tab that TA names are shown in the rows of the table as expected
- This is not testing my changes, but please check to make sure that I did not miss any logic in CalendarDateItem or ProfessorDelete that I should have moved but missed
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorCalendarRow.tsx
+++ b/src/components/includes/ProfessorCalendarRow.tsx
@@ -3,17 +3,7 @@ import moment from 'moment';
 import { Icon, DropdownItemProps } from 'semantic-ui-react';
 import 'react-datepicker/dist/react-datepicker.css';
 import ProfessorOHInfo from './ProfessorOHInfo';
-import { firestore } from '../../firebase';
-
-const getUsers = async (sessions: FireSession[]): Promise<FireUser[]> => {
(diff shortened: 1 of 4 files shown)


ngwattcos (approved):
LGTM - tested locally and it works, all session-related firebase logic has been factored out!

PR #558, opened 2021-05-06
Final changes

pooja654:
### Summary 

 

 Changed a tags to div tags to facilitate single-page app. Created calls to history.push instead of using hrefs. 

 Changes CalendarHeader.scss file to have formatting for div tags instead of a tags.

### Test Plan 

 Tested using yarn lint, yarn start. Manually tested features by checking each option on the menu and making sure it does not lead to another page (it is a single-page app still). When you toggle, it should correctly switch to the proper view. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


pooja654:
Build failed

Issue #478, opened 2021-01-27
Update Queue/Dashboard Toggle to Design + Animated


ngwattcos:
Update from dropdown to toggle:

https://www.figma.com/file/h0kMHuGjX9QCtZ0Y6rUbDj/Fall-2020?node-id=709%3A143

Can use third-party pre-built tools like:
https://www.npmjs.com/package/react-switch

Or have better control over styling with custom-made toggle.

PR #560, opened 2021-05-08
Change Queue/Dashboard to toggle

nada-attia:
### Summary 

Addresses #478 

Changes the display of the dashboard/queue options on the professor's side from a pop up menu to a toggle bar to match the final top bar design from last semester which can be found [here](https://www.figma.com/file/h0kMHuGjX9QCtZ0Y6rUbDj/Fall-2020?node-id=709%3A143).

### Test Plan 

Sign in as a professor, toggle between queue and dashboard, make sure you are redirected to the right page. 
Also, make sure the top bar looks decent on mobile devices. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorStudentToggle.tsx
+++ b/src/components/includes/ProfessorStudentToggle.tsx
@@ -1,6 +1,4 @@
 import * as React from 'react';
-import { useHistory } from 'react-router';
-import Toggle from '../../media/Toggle.svg'; 
 
 type Props = {
     courseId: string;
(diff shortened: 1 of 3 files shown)


triple-jay (changes requested):
Great work Nada! I tested your changes and toggling does direct me to the correct page. I also looked to check how it would look on mobile through responsive design mode and the toggle bar looks good as well. If you don't mind, however, could you make a quick styling fix for mobile devices? As you can see, the professor side bar seems to be somewhat covering the toggle bar. I think you just need to go into ProfessorSidebar.scss and make it so that the top should be slightly adjusted for mobile devices. Other than that, however, great work!

pr_screenshot

triple-jay (approved):
It looks great now :)!

PR #572, opened 2021-08-25
Fixing assignment tags bug

triple-jay:
### Summary 
This should fix professors not being able to see assignments when adding a new assignment, as well as making sure the assignments display correctly, with assignment as the name and tags as provided.

### Test Plan 
Make sure that you can create an assignment and all the fields are correct, including assignment name and tags.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


ngwattcos (approved):
Great catch, great fix. Observed you testing in front of me and it works!

PR #577, opened 2021-09-22
Change all professor/ta email inputs to lowercase

nada-attia:
### Summary 

Change email inputs to lowercase to avoid issues when adding new professors/tas. Previously, if the ta/professor email was all uppercase, a new pending user would be created  in firestore with an all uppercase email. 

### Test Plan 

Add a new professor/ta with an uppercase email. A new pending user should not be created with an all uppercase email in firestore.
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
Thanks a lot for making these changes! I tested with both entering emails as well as uploading CSV files for both professor and TA roles, and it looks good from both the user-facing side as well as Cloud Firestore.

PR #578, opened 2021-10-05
Disabling "@cornell.edu" requirement for test database

abagel21:
### Summary 

Adds a production/staging check to the Cornell email requirement. Fixes a series of eslint proptypes-related issues. As a result of prettier, formatted a bunch of files.

Utilizing the process.env.NODE_ENV and process.env.REACT_APP_IS_STAGING variables, checks whether the current environment is production, and whether the app is not in staging. If this is the case, then the "cornell.edu" header is required. Otherwise, any gmail can be used. Worth noting that the only way to confirm this 100% works is to push to production, but the required state for the header to be active is identical to the check for using the production database in firebase.ts, so any case in which the production database is in use should have the header active.    
For the proptypes fixes, if the type was an object or string, the defaultProp was set to undefined (which should satisfy the undefined checks used previously). The booleans were set according to their use, and what the expected behavior is in the default case.  
The style changes are as a result of prettier (now disabled).

### Test Plan 

I went through connecting with a new gmail account, adding/removing courses, creating office hours, editing office hours, creating and responding to questions, and the admin backend for switching roles.

### Notes 

I added a @eslint-disable-next-line for handleJoinClick in the Props of SessionQuestionsContainer because eslint was complaining about it, but commenting it out triggers a chain of errors in components dependent on it and because I was unsure of the exact behavior of this function, I preferred to ignore the linting error over causing undefined behavior.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
Thank you so much for making these changes Alex! This should really help speed up our development in the future and make it **so much easier** to test new features. Even though we can't test this on production yet, I would 100% agree with you that this shouldn't cause issues, as we have used the same check on the environment variables NODE_ENV and REACT_APP_IS_STAGING in firebase.ts way before this PR. I made sure that the Google Sign In method worked for non-Cornell emails, and I tried adding a question as well. Thank you so much once again for making this change!

PR #579, opened 2021-10-06
Merging into Release

triple-jay:
### Summary 
I'm merging into release because of the importance of Alex's new PR, which allows us developers to log in with normal Google emails onto the test site and is extremely important for testing.

### Test Plan 
Make sure the fix on lowercase emails for adding TAs/professors via CSV upload or manual addition in Nada's PR and logging in with normal Google emails in Alex's PR both work. Also, make sure Nada's PR that fixes the tag bug with creating questions without a tag also works.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ _Screenshots showing major parts of app_
 * **Nada Attia** - Software Developer
 * **Pooja Kanumalla** - Software Developer
 * **Alex Nagel** - Software Developer
+* **Dana Kim** - Software Developer
 * **Jeremy Jung** - Technical Project Manager
 * **Hank Lin** - Product Manager / Designer
 * **Chenchen Lu** - Product Manager / Designer


danakim21 (approved):
Looks good!! The code looks cleaner & easier to read with the formatting changes too. 👍

pooja654 (approved):
I tested Alex's new login method with both a Cornell email and non-Cornell email thoroughly, and it worked perfectly. Great job making the testing process so easy. Nada's fix also works - I tested multiple office hours with no tags and all of the questions worked! This will be really useful.

abagel21:
I tested that logging in with different emails works again, that questions can be created without tags (and answered, manipulated, etc as normal), and that varying the case within emails does not affect the cloud functions ability to process pending users upon account creation. Also tested some other basic functionality of the site to ensure the default props do not affect any of the components--moving around the calendar, interacting with the queue, adding OH/discussions, adding/changing TAs, and all behaves normally. Per the logic of the code (same check for production database as for requiring the Cornell email header), the different emails should not work on the production database, but we should verify that once merged to release.

PR #584, opened 2021-10-13
Done Popup

pooja654:
### Summary 

When a TA marks a student's question as done, they have no option to undo this action. This PR fixes this issue by creating a popup to indicate a question has been marked as done. This popup contains an undo button, allowing a TA to undo the question having been marked as done. The popup automatically disappears after a few seconds via setTimeout() and clearTimeout().

### Test Plan 
I tested this by adding a question in the queue, and from the TA/Professor view, marking done and then either:
* Clicking undo (the popup should disappear, and the question should remain)
* Not clicking undo (the popup should automatically disappear)

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
Thanks a lot for making the requested fixes! It looks good to me now :).

PR #587, opened 2021-10-21
Completely new version of comments

triple-jay:
### Summary 
This might be one of my most ambitious PRs ever. This adds the new comments feature for TAs, students, and professors, which completely changes how comments are run.

This PR allows students, TAs, and professors to comment in a thread-like style with one another, facilitating better communication. The specifics of the new features that were added are detailed below.

- Created a completely new type for comments, FireComment, and structured it so that each question contains a comments collection once comments are made
- New reply button, as well as new look for no show, done, assign from me/unassign from me, and remove buttons
- Create a new comment by clicking on the reply button, entering your comment, then clicking post
- Edit and delete comments through use of the buttons (delete is hidden by default and is accessed by clicking on the three dots ... in the comment, as it is a destructive action)
- Edited security rules to allow only TAs/professors/the student asking the question to read the comment and to allow only the user who posted the comment to delete or edit the comment
- Click on the gray bar to the side to collapse comments (i.e. hide them)
- On the TA/professor side, allow TAs/professors to see the names and profile pictures of commenters, and on the student side, allow them to see if a TA posted a comment or if it is their own comment
- Sorted comments so that the most recent comments come up first

### Test Plan 
- Try adding/editing/deleting a comment, both from the professor/TA side and the user side
- Check the Firebase security rules and test them to see if they work
- Check to see if the user profiles/names show up for TAs and just an anonymous profile picture and generalized name (TA or student) for students
- Make sure the no show, done, assign from me, unassign from me, and remove buttons still work

### Breaking Changes 

Firebase security policy change

- [ ] My PR adds a @ts-ignore


danakim21 (changes requested):
Same as what Pooja commented, I can't see the comment from the student's view but I can see them from the professor/TA's side. Also, there seems to be an error for the profile images as shown below. I'm not sure if this is supposed to be like this or there should be another general profile picture. 

Screen Shot 2021-10-31 at 12 42 58 PM

Everything else seems good to me!! Great work 👍

abagel21:
From what I can tell, the issue with comments not showing up seems to be that the component state (with the new comments from the database) doesn't update until the user updates the component by leaving a comment themselves. For example, I leave a comment from the student side, and I cannot see it from the TA side until I leave a comment as the TA, at which point both my new comment and the comment from the student side load.   
Another small thing, which maybe we should talk with the designers about--the delete functionality is under a dropdown, but the edit button is not. Perhaps we should consolidate, because they fall under the same category of UI manipulation, and the dropdown becomes more effective with multiple elements. Regardless, this is a fantastic improvement of the comments!

nada-attia (commented):
i tested the new realtime feature that you just added and i think it works really well for when a user posts a comment. however, when a user edits a comment the changes don't seem to appear in realtime (they only appear when i refresh the page). to fix this, i think you would need to make the same changes for the function that is responsible for editing comments. 

i have a few comments about the design of the feature (these are small things that we can discuss later):
1. is there a reason why the comment, assign to me, and remove buttons now only appear when we hover over them? it took me a while to figure out that i had to hover over the bottom bar to see the comments button. (i think this is more of a design concern)
2. would it be possible to allow both pressing the "post" button and the enter key to post a comment? i think that would make it a lot more intuitive and i think it would be a small update.

 i think u did an amazing job on this really big feature jeremy 😀

pooja654 (commented):
I also tested the new realtime feature, and it looks super clean - great job! I know Nada said she had to refresh to see changed when a user edits a comment, however, I did not have this issue. When I, as a TA, edited a comment, it updated without having to refresh. 

The comments are still not showing up on the student side, but from the TA view, they look good.

I also agree with what Nada said about the hovering - I didn't realize I had to hover. I also still wish the buttons were a little bigger, because they are important and seem a little small right now.

This is really cool, though!

abagel21:
The issue we had discussed seems to be resolved, with the exception of editing as noted by Nada. The state correctly and reactively updates to changes across browsers for adding comments/replies, assigning, removing questions, and removing comments. I tested the same UI features as previously (hovering, expanding and closing comments, navigating between office hours, viewing multiple questions) and all seems to work. Two small UI things--  
- I don't think this is specific to this PR, but it may be worthwhile to add a little bit of padding at the bottom of the screen, and also more padding onto the question components. When there are multiple comments or a long thread, the components are somewhat glued to the bottom of the screen. It's slightly hard to picture, but if you ask questions on two different accounts, the TA view has the following space at the bottom of the screen after scrolling as far down as possible:   
image
The same case occurs between two questions and below the "Unassigned questions" bar, where especially the latter can seem somewhat indistinguishable:   
image

image

- I'm not sure if you changed this, but I believe Chenchen asked that you move the bar a couple pixels to the right? Feel free to correct me if this is not the case.

Sorry for harping on the little bugs and UI things, I know it can be pretty annoying to keep squashing them. Regardless, the changes look great and the extra reactivity really amplifies the feature!

PR #574, opened 2021-08-25
Revert "Adding product updates single display and modal for seeing all product updates"

ngwattcos:
Reverts cornell-dti/office-hours#553 due to bug

PR #533, opened 2021-04-05
Fix style warning in blog post PR

abagel21:
### Summary 

Merged blog post PR into master, then realized there was a style warning that causes compiling to fail.

### Test Plan 

Just a two line style fix (== -> ===, != -> !==) so not really any testing.

### Breaking Changes 

None

- [ ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore

PR #590, opened 2021-10-25
Readding Product Updates and Notification Display

abagel21:
### Summary 

Readds #553 with additional changes to make the styles more mobile-friendly and reactive.

The bullet pointed changes are the same as in #533, with the exception of small changes to the details of how they are implemented and designed.

### Test Plan 

I tested the following:
- Adding a product update and ensuring it appears with a notification mark in the bottom left and in blue
- Editing a product update and ensuring the changes propogate
- Adding a question and checking that TAs receive both web and in-app notifications
- Responding to a question and checking that the asker receives both web and in-app notifications
- Ensuring that when a new notification appears, viewing it causes it to disappear
- Assigning, unassigning, and marking as done--essentially completing the full OH workflow to ensure the changes didn't affect it
- Changing screen size and shape to test that the displays for the two dropdowns and one modal remain readable and correct.

### Notes 

Still not sure what bug Scott mentioned in #574
ALSO IMPORTANT: This change relies on making sure we have product updates within the production database before they are merged into release.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (changes requested):
Thank you so much for making these revisions Alex!! I tested and found that almost everything works, with the one exception being the notification when a student reaches the top of the queue. I'm not seeing this personally when I click on the bell icon. However, with that being said, I'm not sure if you meant that the notification is something that you implemented, or if you were referring to the browser notifications that already exist when a student reaches the top of the queue. If it is the latter, then feel free to let me know, and I can change this to an approve immediately :). Regardless, thank you so much for going above and beyond to make these changes!! These are really big changes that will have a huge impact on Queue Me In---I can already imagine how helpful notifications will be.

nada-attia (commented):
the new notifications feature looks really good @abagel21! one thing: when a TA marks a question as no show and then clicks on undo, the student gets two duplicate notifications saying that "a ta has been assigned to ur question." this can probably be addressed in a follow up PR since its pretty minor.

Screen Shot 2021-10-31 at 12 12 16 PM

abagel21:
Pushed new changes, addressing both comments:
1. Apologies, the top of queue notification was not being added to the database. That is fixed now.
2. To Nada's comment, I have removed the secondary notification for when a TA is unassigned (the check only looked for if the previous answererId !== current asnwererId) but I am not sure how we want to address this behavior. Currently, there is no notification, but do we want to separately notify a student when a TA is no longer addressing their question?

abagel21:
Shifted to removing notifications after 24 hours rather than immediately after viewing, and added a notification for when TAs are unassigned. Discovered a knew issue with what seems like an attempted fix in place that is not working--browser refreshes duplicate notifications.

triple-jay (approved):
Thank you so much for making these changes Alex---I know it must be hard because with huge features, there's a lot more to check to make sure it works correctly. I've checked that the top of the queue notification is showing, as well as that the TA unassigned action only gives one notification. I also tested that notifications are not immediately removed after viewing, with the old ones getting a white color.  Thank you **so much** once again for your hard work on this!! This will bring such big improvements to Queue Me In :)

abagel21:
In order to fix the bug with duplicating notifications (I only noticed it with the "your question is up" and "a new question has been added" notifications) I added localStorage information to store whether a question is up notification has been set and storing all the previous questions in local storage. For the latter, I was careful to make sure that the questions are only stored if the user is a TA or professor, so the information would already be available to them through the site, but this is an additional security concern. Ironically enough, there are similar persistence solutions for redux (redux-persist library) but until redux stores the questions these are something for the future. To test that these bugs are gone I:
- Added a question
- Tested that refreshing on the student side did not duplicate the "your question is up" notification
- Tested that removing the question as a student, then readding another question gave a new "your question is up" notification
- Tested that on the TA side, a "new question added" notification was sent every time a new question was added, but refresh did not duplicate these notifications
- tested the full life cycle of a question sequentially in a row and that the notifications sent normally throughout this process.

triple-jay (approved):
Thanks again for working on this so thoroughly to make sure no bugs make it to production! I made sure that the notifications are not duplicated on either the TA or student side when a new question is added or the student's question is up, respectively, when the page is refreshed. I also made sure the new question is added notification got added after every new question. I also tested that removing a question and adding a new question got the your question is up notification when the question got to the top of the queue. Feel free to merge this PR in if you feel like it's ready now! If you need me to do any additional testing, I will also very happily do that :).

PR #592, opened 2021-10-30
Create first firebase function unit test

nada-attia:
### Summary 

This PR creates the first firebase function unit test.  It tests the `updateCourses` firebase function which is found in `firebasefunctions/courses.ts`. The calls to firebase such as `docs`, `collection`, etc. have been mocked, and the function has been tested given a dummy user. Now, that this test has been created, unit tests for the remaining firebase functions can be created easily because this test can be used as a reference.

### Test Plan 

run `yarn test` and you should see that the `updateCourses` test is passing
### Breaking Changes 

None

- [x] My PR adds a @ts-ignore
I ignored the expected return types of the mocked objects to be able to replace them with simpler types which makes testing the functions easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "@types/enzyme-adapter-react-16": "^1.0.6",
     "@types/mocha": "^8.0.3",
     "@types/node": "^12.12.15",
+    "@types/sinon": "^10.0.6",
     "chai": "^4.2.0",
     "firebase": "^8.0.1",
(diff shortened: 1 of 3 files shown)


triple-jay (approved):
Thanks for working on this Nada! I tested to make sure that the updateCourses test is passing. Make sure to fix the warnings before merging, but other than that, everything looks great!

PR #593, opened 2021-11-12
Admin course creation bugfix

Archit404Error:
### Summary

Fixed bug where term and year values were switched and cleaned up the form itself. Also changed semester from input field to just text, since semester (FA21) is simply the term(FA) + year(21) (it didn't make sense to have the admin re-enter this info). Also added form input validation.

### Test Plan

Created several sample courses and ensured they worked/validated properly.

### Notes

In the future the form itself should be made prettier.

### Breaking Changes 

None

### Checklist

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AdminCourseCreator.tsx
+++ b/src/components/includes/AdminCourseCreator.tsx
@@ -34,8 +34,8 @@ const AdminCourseCreator = ({ onSubmit }: { readonly onSubmit: () => void }) => 
             alert('Incorrect term format: enter FA or SP');
             return;
         }
-        if (isNaN(year as unknown as number)) {
-            alert('Please ensure the year is a number!');
+        if (isNaN(year as unknown as number) || year as unknown as number < 0) {
+            alert('Please ensure the year is a positive integer!');
             return;
         }
         const course: Omit<FireCourse, 'courseId'> = {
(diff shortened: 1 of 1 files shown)


triple-jay (changes requested):
Thank you for making these changes Archit!! I did check that the year and term are now correct on the admin creation page---however, could you get rid of the input validation for just the course id? Usually, we structure our course ids like tc004-fa-21, which is the course code followed by the semester and then the year. Thus, the course ids are not integers. 

Other than that, thanks a lot for working on this!! You're really onboarding very fast :)

Archit404Error:
> Thank you for making these changes Archit!! I did check that the year and term are now correct on the admin creation page---however, could you get rid of the input validation for just the course id? Usually, we structure our course ids like tc004-fa-21, which is the course code followed by the semester and then the year. Thus, the course ids are not integers.
> 
> Other than that, thanks a lot for working on this!! You're really onboarding very fast :)

Got it! Would you like me to check whether the course id is in "tc004-fa-21" format or just remove validation for course id entirely?

triple-jay:
I think removing validation is fine :) it's just going to me mostly the PMs or I on the admin page so I think you're good!

Archit404Error:
Ok sounds good! I removed the course id validation and also just added in some validation to ensure year is a positive integer!

triple-jay (changes requested):
Thanks for making that change! Just one more thing that I noticed:

I noticed that the course code asks for a 4-5 letter department. However, take CS 2110 for example. This takes a 2 letter department. Since it really depends on the department, I think it's fine to remove validation checks for this as well. Once that's done, I'll approve!

Screen Shot 2021-11-14 at 1 28 44 PM

triple-jay (approved):
Thank you so much for working on this Archit!! It all looks good to me now. Make sure to pull from master to fix any potential merge conflicts before merging this branch in!

PR #594, opened 2021-11-22
Add more unit tests 

nada-attia:
### Summary 

Created unit tests for `calendar`, `courseSettings`, and `session` by mocking the database and making sure that the database method calls are being done correctly (pass in the right arguments, and get called once). 

### Test Plan 

run `yarn test`
### Breaking Changes 

None

- [x] My PR adds a @ts-ignore (to simplify the return types of the mocked objects)

### DIFF
--- a/src/test/local/calendar.test.tsx
+++ b/src/test/local/calendar.test.tsx
@@ -1,0 +1,27 @@
+// @ts-nocheck
+
+import { expect } from 'chai';
+import sinon from 'sinon';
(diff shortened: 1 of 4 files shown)


triple-jay (approved):
Thanks for working on this Nada! I can't express how grateful I am that you took the initiative to learn unit testing when no one else on our team has this expertise. I checked that yarn test works as expected. These unit tests will definitely help us a lot in the future to make sure we aren't breaking anything with our changes!

PR #595, opened 2021-11-22
Redux Infrastructure and User State

abagel21:
### Summary 

Adds redux to the application and incorporates the user object into the redux store.

- Adds the redux, react-redux, redux-thunk, and redux-devtools-extension packages
- Adds the redux store wrapper around the app, and initializes the store with devtools and thunk middleware
- Creates the root reducer, currently only storing auth/user state
- Uses the useMyUser() hook to update the redux state upon any changes to the firebase user object in App.tsx
- uses auth.onAuthStateChanged() firebase function to update the authentication status if it changes in App.tsx
- Removes all props drilling of the user state and replaces it with react-redux state-to-props maps.
- Sets up the infrastructure (both actual code and directory-wise) for additional state

To summarize, overall the current plan of attack is to use firebase hooks, initialized only in App.tsx, with useEffects() to trigger redux actions to update state when the firebase documents change. Through this, all state is updated at the top level (although non-hook based updates can be updated separately) and thus we can remove all props drilling by making each piece of state incorporated through react-redux into the component props, and at our discretion we can have small amounts of props dependencies between small groups of components. This PR includes the initial infrastructure and the implementation of redux state for user and authentication state.

### Test Plan 

The important testing is two-fold: 
1. Any time the user state changes as a result of an action by the user or another user, the redux state changes   
   a. Logging in   
   b. Logging out   
   c. Adding/removing courses
   d. Being added/removed as TA/Professor for a course
2. Components that depend on the user state still behave correctly
   a. adding/removing courses changes the view correctly
   b. questions belonging to a user are assigned correctly (QuestionView)
   c. bookmarks for sessions are added correctly, TA status observed correctly (CalendarView)
   d. courses selected/unselected on course selection are displayed correctly
   e. product updates and notifications still time correctly (e.g. display blue when the user has not viewed them)

To test all of these, I went through the whole user cycle of QMI, starting at the landing page, with the Redux Devtools open (Ctrl+Shift+I, select Redux from the dropdown, then select State and monitor the user and auth state). 
- When logged out, user undefined and auth state false.
- logging in correctly initializes user and sets auth state true
- calendar view and topbar view have correct markers (bookmarks, TA/Professor)
- adding courses, adding questions, discussing, and resolving all works correctly with state updating correctly and notifications marked according to time
- logging out sets user to undefined and auth state to false

### Notes 

I chose not to use a package to implement the firebase->redux connection, utilizing the existing rxfire package with the observables and react hooks to forward the data.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


danakim21 (commented):
I've checked that redux state updates when logging in, logging out, adding/removing classes and also checked that the corresponding courses are shown correctly with their roles. Both UPDATE_USER and UPDATE_AUTH seem to work correctly. The code looks much clearer and intuitive now that redux is implemented! I couldn't check if redux state changes when a user is added/removed as a TA/professor for a course because I'm not sure how to do it on my end, so it would be great if someone else checks on this. Besides that, everything looks great to me.

triple-jay (approved):
Thank you **so so much** for working on this Alex!! I can't express how much I appreciate how you independently researched about Redux and made this PR---it really shows how much you care about taking the initiative for the good of our team!

I reviewed to make sure that the Diff in the Redux Dev Tools shows when a user's role is changed to a TA/professor and demoted from that role, with the action type UPDATE_USER, so you should now be good to go :). Just make sure to fix the **merge conflict** before you merge!

This whole Redux PR looks very nice overall, and I really appreciate how much time and effort you put into this! Let me know if there's anything further you want me to test!

PR #597, opened 2021-11-29
Enable Browser Notifs Popup Redone

pooja654:
### Summary 

This PR changes the design of the enable browser notifications popup. It includes a Got It button, which when clicked, dismisses the popup.

### Test Plan 
To test, I clear my search history, and then log in to QMI as a student. Then, this popup should show up once you click on an office hours at the bottom of the screen. It stays until you hit "Got it," and then it disappears.

### Notes 

There was no requirement to do so, but I was wondering if we wanted this to go away after a bit? Or do we want this popup to only go away when a student hits Got it? Personally, I think it should stay until a student hits Got It so we ensure they see the popup.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (changes requested):
Thanks so much for working on this Pooja! I did test to make sure that the popup shows up as expected. As a very minor fix, could you make the cursor turn into a pointer when you hover over Got It, to make it clear that it's clickable? To accomplish this, all you need to do is add a cursor: pointer line in the CSS.

I agree with what you said that it should only disappear once a student clicks it. I think that if it disappears, they might forget to enable the notifications. 

Also, this is just something that I want you to think about and something that we should also bring up at the standup. Currently in your PR there is also a large notification modal that shows up telling you to enable browser notifications (it has always been there in addition to the popup you changed). Do you think we should keep it, or should we remove it, since it's a duplicate? I attached a screenshot so you know what I'm talking about:

Screen Shot 2021-11-29 at 5 09 08 PM

In any case, great work on this! I'm totally fine with approving as soon as you just add the cursor line to your CSS to make it a pointer.

pooja654:
Jeremy, 

I just fixed the cursor! Thanks for pointing that out - definitely needed to be fixed.

In regards to the duplicate popup, I noticed that as well and was thinking we could take out the one you screenshotted above. I definitely think it's worth bringing up in standup!

triple-jay (approved):
Perfect thanks a lot!! It looks good to go now :)

PR #599, opened 2021-12-05
Fixing plenty of bugs and adding redux state for course and session

abagel21:
##### UPDATED TEXT FOR PR:
After discovered that no notifications were able to be sent when the user is not currently on the course and session wherein the trigger for the notification occurs, I realized that it was necessary to migrate the notifications to firebase cloud functions in order to create this behavior. This is because the client only has access to session data they have loaded, so they are unable to compare changes against session data they don't have. This PR now, in addition to the non-notification related bug fixes and redux inclusion of session data, also includes the changes required to migrate all of the notifications to firebase cloud functions. This includes
- New onSessionUpdate and updated onQuestionCreate and onQuestionUpdate firebase functions
- Removing all notification logic from SessionView and SessionQuestion
- Adding logic for pulling notifications from firebase and sending client-side notifications (browser or eventually text) for new notifications
- Adding a lastSent timestamp to the notificationTracker so that the client side can compare when notifications were last sent with the createdAt field of each notification
- Added a wasNotified marker to each question so that firebase functions can observe whether a user has been notified that their question has reached the top of the queue.

In order to redeploy the firebase functions, I also had to update the yarn.lock for the functions sub-project.

##### ORIGINAL TEXT OF PR:
I discovered a bug with changing classes while having an office hour highlighted--the session hook does not change even though the props to SessionQuestionsContainer does. This essentially means that a user who switches from a class they are a student in to a class they are a TA in will still be viewing the session for the class they are a student in, but will now have the TA role because the course displayed has changed. This secondarily caused a significant number of notifications to be outputted. I fixed this bug by adding redux state for sessions, and ensuring that the session state is set to "undefined" whenever the course changes. Redux now stores course and session state, and passes it down (even though there was not particularly much props drilling in this case, it was helpful still for other reasons, particularly this bug).

In the process of fixing this one bug, I discovered several others. One occurs where asking a question and then switching courses sends the "question marked as complete" notification because the component unmounts. There are three cases when the question component unmounts:
1. The user removed the question themselves. This sets the index to -1 (even though the status is not yet updated) and thus the "question removed" notification is now sent
2. A TA marks the question as done/no-show. This only occurs when the status is 'assigned', and the index is also set to -1.
3. The component is unmounted because the user navigated away--this can occur at any status, but the question index will never be -1 for a question still on the queue.   
I fixed this bug by simply checking that the index is -1 when unmounting before sending any notification, which can only coincide with a non-unresolved state when the question has been marked as complete by a TA.

Another occurs when asking a question, and then switching between office hour sessions, duplicating the "your question is up" notification. A side effect of the attempted solution to refreshing duplication also causes it so that if you ask a question in a different session while already having a question up, it will not send another notification. I fixed this by making the localstorage of questionUp notification status list by questionID, so multiple questions can be monitored, and limiting the ways that the questionUp notification flag for a question can be removed.

Another bug occurred with notifications being sent when switching between past sessions (notifications did not check if sessions were past). This one was fixed by just checking that the session is current before sending any notifications.

On the TA side, the "A new question has been added" notification was duplicated when switching between office hour sessions. I added a prevSession data point to localStorage, and checked that the prevSession was the same as the current session before sending a notification (so that switching sessions does not trigger notifications).

### Test Plan 
The old testing plan, other than the non-notification-related bug, was too specific. The new changes overhaul the entire notification system, so it is necessary to test that every notification behaves as expected.

Switching courses TA view bug
- Add a question
- Navigate to another course (any course, but testing with one where you have a role is ideal because that is where the bug originated)
- Ensure that no session is selected/visible when this occurs

I recommend testing these notifications normally (with the session open), with the user who should receive the notification on a different session, and with the user who should receive the notification on a different course. The behavior should be the same, so it's not necessary to do this with every single one.
- Your question is up (both for adding to an empty queue and marking as done/removing a question ahead)
- A new question has been added
- New student comment
- New TA comment
- TA assigned
- TA unassigned
- question marked as resolved
- question marked as no show

Specifically for the "A new question has been added" notification, I recommend testing the behavior when TA/professor is logged out and a new question is added. As I explained at standup, at the moment the expected behavior is that when a user logs in, all their notifications should be cleared, but if they already have a cookie this will not be the case.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (changes requested):
Thank you so so so much for doing this Alex---this is by far the most thorough and well-written PR that I have ever seen!! The amount of dedication and commitment you show to this team should really be held as a gold standard for everyone on this team, including me! I can't express how grateful I am that you went out of your way to fix **so many** important bugs for Queue Me In. 

Here is my feedback based on what I tested:

- I made sure that the switching courses TA view bug no longer exists---great work on this!!
- The switching courses notification bug is fixed, BUT under the condition that I'm understanding your PR correctly---does the "remove" notification just mean the marked as complete notification? If so, that's what I got, but I wasn't sure if that was the one you were talking about
- I found an additional bug that I think is **very important** that we fix: if you are on another course or another office hour when a TA gets assigned to your question or does anything else (e.g. your question is marked as done, a no show) this notification is **not sent**, even when you return to the question and specific office hour session where it was supposed to be sent
- I tested to make sure the "your question is up" notification does work, but I wasn't sure exactly what you meant by active and inactive courses, as we aren't able to access inactive courses from past semesters
- The past session notifications work as expected :)

Overall, great great work on this!! The fact that you were able to accomplish this in 2 days is what is most impressive to me :).

triple-jay (approved):
Thanks for working on this Alex! I tested now to make sure that the notification bug related to a notification not getting sent if you're on a different course no longer happens, as well as that when a question is removed no notification is now sent. Great work on this and thank you so much for your hard work!

PR #605, opened 2022-02-01
Fixing bug in cloud function onUserCreate

abagel21:
### Summary 

Jeremy noted at the beginning of the semester that there was an issue with adding users by CSVs such that roles were added to the user object but the users were not added to the role arrays for the course, thus preventing them from doing a lot of restricted actions bound by security rules, such as adding tags.

I located two issues in the onUserCreate function:
1. The mapping function for courses where the user is a professor used taCourseIds instead of profCourseIds, which meant that the course was always undefined and the user was never added
2. When adding users to the array for the course, the array was set such that it overwrote all existing data (tas= [userId]).
  
Based on my testing, these have both been rectified. 
  
IMPORTANT NOTE: I have pushed the changed function to QMI-test, but not prod. I'm not sure if we have to separately deploy the functions to prod after this PR is merged.

### Test Plan 

Because the scope of this bug is severe, I wanted to make sure that all ways of adding users to roles work.
- Adding an existing user by CSV/form
- Adding a nonexistent user by CSV/form

For the latter, I logged out of an account, deleted the user object from the firebase database, and deleted it from the role array in the course I was targeting (if necessary). Then, I added via CSV such that the user becomes a pendingUser as a professor and then a TA (testing both). 

Regardless of whether a preexisting user or not, I verified that the role appeared both under the user object and the userId appeared under the course array for that role.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -10,7 +10,7 @@
     "logs": "firebase functions:log"
   },
   "engines": {
-    "node": "10"
+    "node": "12"
   },
   "main": "lib/index.js",
   "dependencies": {
--- a/functions/package.json
+++ b/functions/package.json
@@ -10,7 +10,7 @@
     "logs": "firebase functions:log"
   },
   "engines": {
-    "node": "12"
+    "node": "10"
   },
   "main": "lib/index.js",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


triple-jay (approved):
Thank you SO SO MUCH for making these Alex! I checked to make sure with both cases that when a non-pending user or a pending user is added (and in the latter case, the pending user joins QMI), the TA and professor list is as expected for the course collection and the property in the user collection is as expected.

I can't believe this bug went undetected for so long---it makes complete sense why our system was acting in that way of having only one TA on the list. THANK YOU SO MUCH once again!!

PR #606, opened 2022-02-01
Fixed CSV bug

triple-jay:
### Summary 
Fixed a CSV upload bug where uploading a CSV that updated the roles of users would not entirely work.

I noticed that if you updated the roles of users in both the professor and TA roles, the updates would not be correctly reflected in the admin page---the courses collection would not have the correct list of TAs and professors.

### Test Plan 

- Have a CSV file that updates the role of some users as TAs and some users as Professors (MAKE SURE you have both), and proceed with import
- Make sure the admin page and the Firestore database shows the correct user placement in both the users collection in the roles field as well as in the courses collection in the tas and professors fields.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
I tested with a CSV upgrading one student to a Professor and another to a TA, and it correctly set the user roles and the course role arrays in the firebase database. I also checked the admin page, where the changes were similarly accurately reflected. Nice work!

PR #607, opened 2022-02-01
Important bug fix for importing TAs/professors and pending users

triple-jay:
### Summary 
Pushes very important bug fixes to production

### Test Plan 
- Make sure importing professors/TAs works as desired, and try testing pending users to see if when a pending user creates an account, the list of TAs and professors is still as expected (see #605 and #606)

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
I've tested both PRs and they function as expected. LGTM

PR #613, opened 2022-02-22
To release

triple-jay:
### Summary 
Adding calendar export and necessary bug fix to release

### Test Plan 

- Make sure calendar export works
- Make sure no duplicate notifications are sent in any case
- Make sure when switching to a different class the office hour session does not persist
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


pooja654 (approved):
I tested both calendar exports - Apple and Google - and both seem to work fine. I asked a question as a student and switched between three different classes successfully without any issues. When I came back, my question was still there. I also made sure that I received a single notification (not multiple) when it was my turn in the queue, and this worked as well!

PR #614, opened 2022-02-22
change color

erin-xu:
### Summary 

Changed ahead to red and finished to green.

### Test Plan 
- check the colors

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
I checked that the finished text is in green and the ahead text is in red! Great work!

PR #616, opened 2022-03-02
Added time limit for questions

triple-jay:
### Summary 
Adding a time limit option that professors and TAs can enable so that a sound and banner will display at a warning time and at the time limit time for each question, signifying that the TA should move on to the next question.

I added an option in the professor dashboard under the settings button to add a time limit as well as a time warning for questions, such that at those times, a banner will display and a sound will play, letting the TA and student know that the time for their question is almost up or finished. This can be disabled in case the professor does not want to use this.

### Test Plan 
- Go to the professor view and go to "settings" and enter a time limit as well as a time warning for a class
- Add a question as a student and assign yourself as a TA, and wait to see if you get the banner and sound for the time warning you specified as well as the time limit
- Due to bugs and also just because it was something I wanted to add, make sure that a student can't remove their question when they are assigned
- Make sure that if a TA unassigns himself or herself, as well as marks as done or no show a question, the banner as well as the sound is not sent
- Try disabling the time limit and time warning with the checkbox and see if the banner and sound are still sent

### Breaking Changes 

- [ ] My PR adds a @ts-ignore


abagel21 (changes requested):
I tested the things you mentioned. The jingle sounds great!      
Entering and updating (changing the value, disabling) the time limit/warning works as expected. I get a warning and jingle, as well as the "time is up" banner at the expected times. Students can no longer remove their questions when assigned as well.   
I have found one issue, however. On the TA side, when marking as done, no show, or unassigning, regardless of before or after the warning, the banner and jingle are both disabled. However, I am still getting both banner and jingle on the student side no matter when these events occur, for all three of these TA interactions.

erin-xu (changes requested):
I followed the test plan and noticed the same issue that Alex mentioned. Everything else worked as expected. I was able to get the banner and jingle on both the TA and student sides. The banner and jingle are not sent when the time limit is disabled, and students cannot remove questions after being assigned.

abagel21 (approved):
I retested the whole workflow, from enabling to disabling, and the banner and sound work as expected. All three actionable cancellations (done, no-show, I don't know) prevent the banner and sound from displaying both before and after the warning. I would recommend removing the [.25, .5] option before pushing to master however.

erin-xu (approved):
The banner and jingle now work as expected, and the student does not receive them if their question is marked done, unassigned or no show!

PR #618, opened 2022-03-03
Added tooltip

erin-xu:
### Summary 
Added tooltip with queue position.

### Test Plan 
- Hover over queue position and check that tooltip with correct position is displayed.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -305,7 +305,9 @@ class SessionQuestion extends React.Component<Props, State> {
                 <div className="TopBar">
                     {!this.props.includeRemove && includeBookmark && <div className="Bookmark" />}
                     <div className="OrderTooltip">
-                        <span className="TooltipText">You are position {this.props.index + 1} in the queue.</span>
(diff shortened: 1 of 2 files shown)


triple-jay (changes requested):
Thanks for working on this Erin! I did verify to see that the tooltip is shown when hovering over queue position :). If it isn't too much work, could you disable this on the TA side? Currently, they can see the tooltip when hovering over each queue position of students, and I don't think they need to see this on their end.

Screen Shot 2022-03-03 at 11 06 20 AM

danakim21 (changes requested):
Great work done Erin!! I also agree with Jeremy's point that the professors or TAs don't need to see the tooltip. Also, it would be better if the tooltip is closer to the question because it seems like there's a lot of space in between right now.

triple-jay (approved):
Looks great Erin! I can no longer see the position on the TA side anymore :)

danakim21 (approved):
Thanks for making the changes Erin!! I've checked that the tooltip is visible only to the students and the spacing is better 👍

PR #621, opened 2022-03-09
Pooja/wait

pooja654:
### Summary 

I updated logos in the session header for students ahead, TAs helping, and estimated wait time. I implemented the feature that lets users know how many TAs are currently helping students in this office hour. I also added a feature that tells a user the time they will be first in the queue (and added it onto the current estimated waiting time).

### Test Plan 

I tested office hour sessions with multiple TAs and singular TAs and the TA count worked. I checked the time calculation multiple different times, and all worked.

### Notes 

Twice, when switching from Queue to Dashboard, I received an error. When I refreshed the page, the error went away. I was never able to reproduce this error - please check this when you test to make sure switching between Queue and Dashboard work!

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (changes requested):
Great work Pooja! I'm really proud of how you've grown while on the team, and this PR really shows that :). Just a very small thing, but could you format the time as HH:MM? I attached the screenshot below to explain what I'm talking about.

Screen Shot 2022-03-09 at 10 07 40 AM

pooja654:
I just fixed the minutes error you pointed out, thanks Jeremy. I also noticed it was having trouble when the waiting time was over 24 hours, so I added in some code to take care of that as well though this case probably wouldn't arise realistically.

triple-jay (approved):
Thank you for fixing this Pooja! It looks good to me now :)

erin-xu (approved):
Pooja, this looks great! I tested office hours with different numbers of TAs, and the TA count worked. I also tested the estimated wait time with different queue lengths and it worked as well. The HH:MM time format that Jeremy mention is fixed too.

PR #623, opened 2022-03-10
Added new type of hybrid office hour with TA zoom link

triple-jay:
### Summary 

I added an option when creating a hybrid office hour that allows the course to provide the zoom link. This is represented by the main office hours screen which tells the student to go to the course to find the zoom link. 

You can create this new type of office hour by checking the "Use course zoom link" box on the hybrid office hour create screen as a professor role.

### Test Plan 
- Test in-person, virtual, and both types of hybrid office hours (course provided zoom link checked and unchecked) and **_make sure_** that they work. You can do this by adding a question as a student for all types of office hours, as I did change the state machine a little bit for this PR, as well as try assigning and answering questions as a TA
- Test hybrid with both series as well as for an individual session for the course zoom link version
- Try checking and unchecking the virtual box for the new type of hybrid office hour to make sure that the label is as expected

### Breaking Changes 

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AddQuestion.tsx
+++ b/src/components/includes/AddQuestion.tsx
@@ -114,7 +114,11 @@ const AddQuestion = ({ course, session, mobileBreakpoint }: Props) => {
             setStage(LOCATION_INPUTTED);
             setSelectedSecondary(tag);
         } else {
-            setStage(SECONDARY_SELECTED);
+            if (session.modality === 'hybrid' && typeof session.useTALink !== 'undefined' && session.useTALink) {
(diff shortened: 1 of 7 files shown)


abagel21 (approved):
I created all 4 types of office hours, and added two questions from two accounts to each queue. The normal workflow (assign, mark done) works, and the other interactions (no-show, I don't really know) also work. I tested with mixed questions (one in person one virtual) for both hybrid queues and they displayed correctly and worked normally, with the student zoom link displaying for the student-provided one and not for the course-provided one, with the requisite messages in the session information header also correct.

PR #624, opened 2022-03-10
To production

triple-jay:
### Summary 

Gets time limit, hybrid badges, and the new hybrid office hour format into production!

### Test Plan 
- Make sure the time limit, badges, and new hybrid format work as expected

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
I've previously tested these 3 PRs and they work as expected.

PR #625, opened 2022-03-11
Very minor fix for location bug

triple-jay:
### Summary 
Allows students to add location in the case where there is no tags.

### Test Plan 
- Make sure you can add questions as expected for all three (four, including the new hybrid type) office hour types without any tags
- Make sure you can see the checkbox for the new type of office hour without any tags

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
Tested same things as previous PR (and noted in this PR) on a course with no tags, all worked as expected.

PR #627, opened 2022-03-11
Added fix for having no tags under assignment

triple-jay:
### Summary 
When creating or editing an assignment, you must have a tag now. A warning message in red text shows up if you try to create or edit an assignment so that it no longer has any tags.

### Test Plan 
- Test to see that you can not create an assignment without a tag, or edit an assignment so that it no longer has any tags
- Test to see that you see the red warning message pop up when doing so

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
Tested adding tags and it made sure I had a subtag.

PR #628, opened 2022-03-15
Updates on average wait time

danakim21:
### Summary 

This PR updates the way we calculate average wait time. Currently, QMI shows the same average wait time for all students regardless of their position in the queue, and students' wait times are considered to have the same weight when we calculate the total wait time of the session. I've changed the way we calculate the average wait time based on this [detailed solution](https://docs.google.com/document/d/1KJhXQdAfdDP_7PQ_5QxpV46ShoCxPvJ4yhZr2ICBWbo/edit?usp=sharing). The major changes are: 

- Added a new optional field called `position` under `FireQuestion` interface 
- Changed `waitTimeChange` to take into account the student's position in the queue 
- Changed estimated wait time to take into account the student's position in the queue by calculating the dynamic position. If the student does not have their question in the queue, their position is considered as _1 + last student's position_. TAs will see the average wait time of the entire session. 

### Test Plan 

- Try adding multiple questions to a session and check that students have different average wait times based on their position. 
- Check that the new `position` field does not cause an error

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


pooja654 (commented):
Dana, this looks great! I tested this with 4 accounts - one professor, three students. When I had 2 students, it didn't test much as one of them showed the same wait time the prof/TA side showed. So, I added a 3rd student to really check that the prof/TA side and student side showed different times, and they did - the student time being larger, of course. 

One thing I noticed that may or may not have to do with your PR specifically is that it wouldn't let me submit a question with no tags...there were 2 assignment options, and one of them had a tag and the other one didn't. For the one that didn't, I couldn't submit a question, so I used the one that had a tag and was able to. Jeremy might have to answer this part. 

Otherwise, great job - this is definitely a helpful PR!

triple-jay (approved):
Dana this is amazing!! Thank you so much for working so hard on this :)

I tested to make sure that the position is dynamic and changes whenever you move up in the queue, as well as that the wait time gives the expected waiting time if you were to enter the queue just at this moment for people who haven't entered a question yet. I also made sure that the TA sees that the wait time is the average wait time for each person moving up one position in the queue. Finally, I tested that adding questions on the test site does not create any issues on the local.

To answer @pooja654, the error that you see with the tags has been fixed last week by me and Alex in that TAs and professors are now no longer able to create or edit assignments so that they have no subtags, so we shouldn't be having that issue anymore.

PR #629, opened 2022-03-18
Removing undefined error (hopefully)

abagel21:
### Summary 

Adding additional checks to notificationList to avoid undefined error being thrown

### Test Plan 

On session view, click on the notification bell. Then, tab between session view and dashboard a couple times. Hopefully no error thrown. The other error I have not been able to reproduce but Michelle can test out opening the test site once we push.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
I did toggle as you said and made sure that it isn't returning any errors! Great job Alex and thank you for getting this done so quickly :)

PR #630, opened 2022-03-19
Calendar day export

erin-xu:
### Summary 
This PR adds a feature to export an entire day of office hours to Apple Calendar. Above the list of office hours on the left side of the screen, I added the date and a calendar icon to export the office hours for the day. Clicking the icon should open the calendar export modal, where the user can choose to export the office hours to Apple Calendar. I could not find a way to export an entire day of events to Google Calendar, so only the Apple Calendar option is shown. I also updated the title of the calendar export modal for a single office hour session to include the course name and OH name. 

### Test Plan 
- Check that the correct date is displayed for the selected day of office hours. 
- Try adding multiple office hours to the same day and exporting them at the same time using the day export feature (icon next to the date).
- Only the Apple option should be shown when exporting a day of OH. 
- Check that the correct course and OH name is displayed in the calendar export modal when exporting a single OH.

### Notes 
I noticed that there are some office hours that start before 12:00am or end after 11:59pm. The Apple Calendar export does not work for those office hours. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CalendarExportModal.tsx
+++ b/src/components/includes/CalendarExportModal.tsx
@@ -48,6 +48,30 @@ const CalendarExportModal = ({
         return { date, startTime, endTime };
     }
 
+    const getDateString = (): string => {
+        const days = ['Sunday','Monday','Tuesday','Wednesday','Thursday','Friday','Saturday'];
+        const months = [
+            'January',
+            'February',
+            'March',
+            'April',
+            'May',
+            'June',
+            'July',
+            'August',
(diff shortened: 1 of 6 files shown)


triple-jay (approved):
Thanks for working on this Erin! I tested that all 3 modalities of office hours are able to be exported as expected, and I also checked that all the information looks right.

burninglilies:
This looks really good!

PR #632, opened 2022-03-22
Adding a <secret> email and password login

abagel21:
### Summary 

At the bottom of the landing page, adds a "Having trouble logging in?" hotlink that opens a email/password login modal. Adds support for adding a user by email/password.

### Test Plan 

Create an account in firebase auth. Log in with that user. Add a first/last name to the user in the firebase database. Test that it can interact with the site normally.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
Thanks a lot for working on this Alex! This is definitely much needed for future professors and TAs that don't have Google emails :)

I did test the process of creating a user in auth and then creating a user document for that user with a first name and last name, and I checked to make sure that there were no problems for that user when the user was promoted to the professor role, such as assigning office hours and creating them.

PR #633, opened 2022-03-27
No Show Popup Modifications

pooja654:
### Summary 

I changed the design of the No Show popup to match Chenchen's new design. I additionally modified the timeout so that the popup lasts longer (for both the no show and done popups). As of now, the move to the bottom of the queue button simply removes the student from the queue. I am planning on changing this to move the student to the bottom of the queue, but running into a few troubles.

### Test Plan 
I tested that both the no show and done buttons work as expected, and last long enough to be read.
I tested that the undo button for both the no show and done popups leaves the student in the queue in their original position.
I will test that the move to the bottom of queue button moves a student to the bottom of the queue, but for now, I've tested that it removes the student from the queue.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AddQuestion.tsx
+++ b/src/components/includes/AddQuestion.tsx
@@ -305,7 +305,7 @@ const AddQuestion = ({ course, session, mobileBreakpoint }: Props) => {
                                             ))
                                     ) : (
                                         <p className='placeHolder'>
-                                            {activeTags.length > 0 ? 'First select a category' :''}
+                                            {activeTags.length > 0 ? 'First select a category' : ''}
                                         </p>
                                     )}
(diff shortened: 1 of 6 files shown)


danakim21:
I checked that pressing "no show" or "done" both show the undo popup and pressing "undo" leaves the student in their original position. As you said, right now pressing “move to the bottom of the queue” is simply removing the student from the queue -- adding this feature would be great for the students!!

erin-xu (commented):
This looks great so far! I checked that the no show and done buttons open the pop up as expected. The pop up is shown long enough to be read, and the undo button leaves the student's position unchanged.

PR #637, opened 2022-03-29
Hide TAs assigned for office hours with no TAs assigned

triple-jay:
### Summary 
Hides "TAs assigned" for each office hour for office hours without TAs assigned

### Test Plan 
- Check that TAs assigned shows up for office hours with TAs assigned
- Check that this is hidden if no TAs are assigned 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
Verified that the line shows if TAs are assigned, and doesn't if they aren't.

PR #642, opened 2022-03-30
Updating master with firebase function changes

abagel21:
### Summary 

Last night, I deployed several updates to the firebase functions to prod to fix a bug. This is a PR to make sure those changes are  on github as well as on firebase.

- Added .catch()es to every notificationTracker update so that users without notificationTrackers can create them and won't throw an error
- refactored some code

### Test Plan 

- Delete your notificationTracker in the database, and ensure that notifications send normally
- ensure that notifications send normally with a notificationTracker too

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -1,6 +1,6 @@
 import * as functions from 'firebase-functions';
 import * as admin from 'firebase-admin';
-import { Twilio } from "twilio";
+import { Twilio } from 'twilio';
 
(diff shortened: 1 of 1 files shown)


triple-jay (approved):
I tested that notifications work with or without a notificationTracker :). Thank you so so much Alex for your hard work last night <3

PR #645, opened 2022-04-14
Fix waiting time calculation

danakim21:
### Summary 

This PR fixes the waiting time calculation, which was previously added in #628. Previously, person number 1 on the queue saw the normal average wait time, but now they see 0s. 

### Test Plan 

- [ ] Check that student 1 in queue has 0s waiting time 
- [ ] TA/professor waiting time should remain unchanged 
- [ ] Students in the queue should have incremental wait time depending on their position 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionInformationHeader.tsx
+++ b/src/components/includes/SessionInformationHeader.tsx
@@ -104,12 +104,12 @@ const SessionInformationHeader = ({
     let dynamicPosition = questions.findIndex(question => question.askerId === myQuestion?.askerId) + 1
 
     if (dynamicPosition === 0) {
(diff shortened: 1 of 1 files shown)


triple-jay (changes requested):
Thanks for working on this Dana! I saw the text was changed on the TA side, as well as that it shows 0s for the first person on the queue now. However, I did see that the calculated time in parentheses for when that student's question would be up is inaccurate—I think it still takes into account one more position (i.e. the extra wait time) in that calculation.

Screen Shot 2022-04-17 at 12 30 34 PM

Other than that, it looks good!

PR #647, opened 2022-04-18
Adding collapsing past office hours, fixing 24 hour time on notifications

abagel21:
### Summary 

Makes past office hours automatically collapsed (unless there are no current office hours) and allows expanding as well. Also makes the time for notifications in 12 hour, not 24 hour. 

### Test Plan 

- Have a course with at least one past office hour and at least one current office hour. On clicking the course (or refresh), the past office hours should be automatically collapsed but able to be expanded.
- Have a course with only past office hours. They should be automatically expanded on clicking on the course (or refreshing), but still able to be collapsed
- check that a notif past noon has 12 hour time, with am/pm correct.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CalendarSessions.tsx
+++ b/src/components/includes/CalendarSessions.tsx
@@ -1,5 +1,6 @@
-import React, { Dispatch, SetStateAction, ReactElement } from 'react';
+import React, { Dispatch, SetStateAction, ReactElement, useState } from 'react';
 import { groupBy } from 'lodash';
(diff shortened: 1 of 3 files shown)


triple-jay (approved):
Thanks for working on this Alex! I tested that the past office hours are collapsed as well as that the time is in am/pm for notifications.

futurekim (approved):
I tested that the collapse is working as well as the am/pm on the notifications :) Looks good!

PR #648, opened 2022-05-01
TA Announcements

danakim21:
### Summary 

This PR adds the new feature of adding TA announcements. For each session, TAs can add new announcements -- if they click on the chevron button, the section expands or collapses. When the section is expanded, it will either show the currently added announcements or "No announcements yet. Click on "plus" icon to make one." When a user clicks on the plus icon, the section also expands and shows a text box for TAs to input their announcement. When a TA/professor adds a TA announcement, it should show to the corresponding session. Adding a new announcement should only work for TA/professors. Everyone should be able to view the announcements. TA/professors can delete only the announcements they created. 

### Test Plan 

- [ ] The chevron button works as expected 
- [ ] The plus button works as expected 
- [ ] The cancel button works as expected 
- [ ] New announcements can be posted by TA/Professors 
- [ ] Previously added announcements show 
- [ ] TA/Professors can delete announcements they've created 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21:
Tested the UI (toggling plus, chevron, different combos of the two) and it all works as expected. One thing to note, in addition to development of the new announcements being added, the name currently displayed at the top of the "new announcement" block is currently the generic one rather than the current user's. Other than that, looks great!

erin-xu (commented):
This looks great so far! I tested the chevron, add, and cancel buttons, and they all work as expected. Other than the default TA name that Alex mentioned, it looks good to me.

pooja654 (commented):
The chevron button, plus button, and cancel button all work as expected. Clicking post pops up the notification, but doesn't do anything else, which I believe is what you said you implemented. I also noticed what Alex said, about the TA name. One more thing is make sure the TA announcements count updates properly - once again, I believe it doesn't now because post isn't fully implemented. Otherwise, great job!

abagel21 (changes requested):
Looks and works great, thanks Dana! Just a couple of nitpicky things--when I first load the page, before any announcements are posted, the grey text upon dropdown doesn't show up and the parenthetical is empty instead of displaying 0:
image
The other tiny thing is that if possible, announcements should display most recent at the top,  instead of like the queue, because we want them to see the most important information first.

abagel21 (approved):
I've retested the workflow and it looks great. Thanks Dana!

PR #667, opened 2022-08-24
Various Fixes from early semester feedback

abagel21:
### Summary 

In discussion with 4410 and 2110, a couple of issues/ambiguities were raised. These were:
- Ambiguity with what "pending" means--it seems like the operation to add users is in progress, but the pending message disappears on refresh, so then it appears it has failed because the names are never added
- inability to add a course link to a session when "use course link" is selected
- Creating a session series without a name fails without a descriptive error message, while creating a session without a name simply works

I've implemented the following solutions, in order:
- Changed the "pending" message to "pending account creation", and added a second table to the manage roles tab that shows which emails are pending and for which roles
- added a course link to the sessions, such that if "use course link" is selected, a box appears to add a link for the session that is then presented to both the TA and the student identically irrespective of office hour state
- added a default title "office hours" so that no office hour is ever created without a name, even if one isn't entered

I also minorly tweaked the notification text,  because there was a newline in the string.

### Test Plan 
1. Add a role to an email that doesn't exist. Check the correct message appears. Verify that the user is added to the pending table. Verify that creating an account with this user removes them from the pending table.
2. Create a session using course links, and test that both TAs and students can access this link. Create a session without course links, and check that the normal TA link behavior still works.
3. create a session, and then a session series, without a name, and verify it works.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -400,7 +400,7 @@ exports.onQuestionUpdate = functions.firestore
                         title: 'TA Unassigned',
                         subtitle: 'TA Unassigned',
                         message:
-                            `A TA has been unassigned from your question and you\'ve \
+                            `A TA has been unassigned from your question and you have \
                   been readded to the top of the ${session.title} queue.`,
                         createdAt: admin.firestore.Timestamp.now()
                     })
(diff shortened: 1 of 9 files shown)


triple-jay (approved):
Thank you for your hard work sir :)

PR #676, opened 2022-10-04
Fixing default values autofilling on every login for manually constructed users

abagel21:
### Summary 

For users that we have manually made a login for (email/password), Google authentication does not supply a name or photoUrl, which results in us constantly updating the database to "Dummy name" and "Dummy photo". I rectified this by making the database transaction only run in the case that we have a defined value for name and photo.

### Test Plan 

Log in normally with your netID and test that your name/phone number/photoUrl are maintained correctly. Log in with a username/password combo (The "having trouble logging in" button at the very bottom right of the home page", can use user:alex@alex.com, pass:alex@alex.com to verify) and check that name/phone number/photoUrl are maintained correctly.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/LoginModal.tsx
+++ b/src/components/includes/LoginModal.tsx
@@ -1,8 +1,7 @@
 import React, { Dispatch, SetStateAction, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { Icon } from 'semantic-ui-react';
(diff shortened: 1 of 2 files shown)


triple-jay (approved):
Thanks for working on this sir!

PR #677, opened 2022-10-07
Fixing default values autofilling on every login for manually constru…

abagel21:
…cted users (#676)

* removing userUpload from loginModal

* fixing dummy name instantiation

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
Looks good sir!

PR #679, opened 2022-10-13
Dana/overlap banners

danakim21:
### Summary 

If both "Enable browser notifications" and "Enable text notifications" banners are up, they overlap so only one banner shows. This PR fixes this issue so that the second banner comes after the first banner. When only one banner appears, it stays in its original position. 

Screen Shot 2022-10-12 at 9 21 32 PM

Screen Shot 2022-10-12 at 9 22 06 PM

Screen Shot 2022-10-12 at 9 22 23 PM

### Test Plan 

I've tested that with two banners showing, they don't overlap. With only one banner, it stays in its original position. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/Banner.tsx
+++ b/src/components/includes/Banner.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {connect} from 'react-redux';
+import { connect } from 'react-redux';
 import { updateTextPrompted } from '../../firebasefunctions/phoneNumber';
 import { removeBanner } from '../../redux/actions/announcements';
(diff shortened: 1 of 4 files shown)


erin-xu (approved):
This looks great, Dana! I checked that the two banners no longer overlap when both displayed. I also checked that when each banner is displayed by itself, it is displayed in the original location.

burninglilies:
Looks great Dana! I tested this on my computer and the banner displays as expected with no overlap

PR #680, opened 2022-10-14
Professor Demotion

erin-xu:
### Summary 

Last year, there were multiple occasions where professors accidentally demoted themselves or other head TAs by including them in their list when updating roles via import. This PR prevents role demotions (professor to TA) when updating roles via import. 

### Test Plan 

I tested the fix by checking that a user cannot be demoted from professor to TA when updating roles via import (both entering manually and uploading csv file). I also made sure that demotions can still be done using the role dropdown option. Additionally, a user's role can still be updated from TA to professor using the import option. 
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/firebase.ts
+++ b/src/firebase.ts
@@ -20,7 +20,7 @@ if (process.env.NODE_ENV === 'production' && process.env.REACT_APP_IS_STAGING !=
     };
 } else {
     firebaseConfig = {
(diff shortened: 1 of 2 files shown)


abagel21 (approved):
I've tested normal promotion, which works as expected, and that demotion fails. I'm going to create an issue for adding a banner that informs the user that demotion failed, but that can be separate from this PR.

PR #681, opened 2022-10-14
fixed alignment for dashboard warning

burninglilies:
**Summary** 

The alignment for the dashboard was misaligned and was squished with the "Please select an office hour from the calendar."
So I made a css style change and changed the display from inline to center text alignment and was able to see a visible improvement after I ran this on my computer. 

**Test Plan**

I've tested that changing the size of the screen would keep the warning in the center.
image

**Notes**

**Breaking Changes**

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/styles/SplitView.scss
+++ b/src/styles/SplitView.scss
@@ -1,6 +1,6 @@
 div.warningArea {
     padding: 0.3in;
-    display: inline-block;
+    text-align: center;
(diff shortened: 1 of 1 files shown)


sw-min (approved):
I checked on my end and the text stays centered throughout resizing the browser!

PR #682, opened 2022-10-27
display fullname

sw-min:
### Summary 

The full name of the user (first name & last name) should now display on the dashboard.

### Test Plan 

I've tested that the first name and last name now are shown in the welcome page. Long names are wrapped to the next line below the welcome statement.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


danakim21 (approved):
I've tested on my side and the full name shows!

burninglilies:
I also tested this on my end as well and it displays as expected!

PR #685, opened 2022-10-28
Live Forum fixes and adding unanswered/answered to normal queue

abagel21:
### Summary 

In discussions, the "Unassigned queue questions" label appeared on top of the live forum UI. This removes that. This also fixes a variety of small visual bugs with live forum, and ports the new comment system/view to it. Long term, these need to be modular components so we don't have to update both every time. It also moves the answered/unanswered toggle to the normal queue (should also be modular).

### Test Plan 
Create a live forum. Verify the unassigned queue questions label does not appear. Go through the whole question workflow with multiple questions. Verify TAs and students can see all questions. Verify upvotes work and change ordering. Go through the question workflow on a normal queue. Verify that students cannot see the toggle, but TAs can and it correctly delineates questions into the correct section.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (commented):
toggle disappears when no questions in queue
assigned queue questions appears on live forum
answered questions displays incorrectly
normal queue question displays after answered

PR #687, opened 2022-10-31
Professor view of all assigned questions

erin-xu:
### Summary 

This PR allows professors to view which TA is assigned to each question. Previously, professors and TAs could only view unassigned questions and questions assigned to them. I added a third section on the professor view that includes all assigned questions, allowing professors to see who is assigned to which question.

Screen Shot 2022-10-30 at 8 51 57 PM

Since there may be a large number of assigned questions, the "All Assigned Questions" section is collapsible to avoid cluttering the session page. 

Screen Shot 2022-10-30 at 8 52 43 PM

When professors assign themselves to a question, this question will be included in the "Assigned to Me" section instead of the "All Assigned Questions" section so professors can easily see which questions they are assigned to. 

Screen Shot 2022-10-30 at 8 52 21 PM

With this new view of all assigned questions, professors can now unassigned TAs from questions. Currently, when a professor does this, only the student is notified that a TA has been unassigned from their question. I still need to add notifications on the TA side.

### Test Plan 

- [ ] Check to see that professors can see a list of all assigned questions and which TAs are assigned to each one
- [ ] Check that the "All Assigned Questions" sections is not visible when there are no other TAs assigned to questions
- [ ] Check that the chevron button collapses and uncollapses the "All Assigned Questions" section as expected
- [ ] Check that questions assigned to a professor are under "Assigned to Me" and not "All Assigned Questions"
- [ ] Check that TAs cannot view questions assigned to other TAs or professors

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -417,7 +417,11 @@ class SessionQuestion extends React.Component<Props, State> {
                                         </p>
                                         :
                                         <p className="Begin" onClick={this.questionDontKnow}>
-                                            Unassign from me
+                                            {answerer &&
+                                                ('Unassign from  ' + (answerer.userId === this.props.myUserId
+                                                    ? 'me'
+                                                    : answerer.firstName + ' '
+                                                    + answerer.lastName))}
                                         </p>
                                     }
                                 </div>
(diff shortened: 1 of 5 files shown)


abagel21 (approved):
I've tested that each section only appears when there is a question that belongs under it, that only professors can view all assigned questions, and the ordering of each section when all three are visible. Questions appear under the correct section when assigned to the current user. Excellent work!

danakim21 (approved):
I checked by going through the test plan and everything works as expected. Great work Erin!

burninglilies:
I checked to make sure the professor views worked as expected and it looks good!

PR #689, opened 2022-11-06
Add logic for pagination to view more questions

danakim21:
### Summary 

This PR adds pagination for TAs to view more questions. Currently there is a limit on the number of questions shown using the constant `NUM_QUESTIONS_SHOWN`, so if there are more questions added, the TAs can't see them. Pagination is added to fix this so that in one page, only `NUM_QUESTIONS_SHOWN` questions are shown at a time. 

### Test Plan 

I've created multiple questions and checked pagination (previous page, next page) works fine. I had to change `NUM_QUESTIONS_SHOWN` to a smaller number to test this. 

### Notes 

Let me know any design recommendations because currently pagination is just done through small arrows! 

Screen Shot 2022-11-06 at 1 51 01 AM

Screen Shot 2022-11-06 at 1 51 06 AM

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestionsContainer.tsx
+++ b/src/components/includes/SessionQuestionsContainer.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useState } from 'react';
 import { Loader } from 'semantic-ui-react';
 import moment from 'moment';
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


erin-xu (commented):
I checked that the arrows navigate the pages as expected. Great work, Dana! 

Since we are now allowing TAs and professors to view questions beyond the first page, they are now able to assign themselves to questions on later pages. I noticed that if a TA assigns themself to a question that is not on the 1st page, the "Assigned Questions" section will also not appear on the 1st page. In the images below, I added 2 questions and set the number of questions shown to 1. I assigned myself to the 2nd question, and "Assigned Questions" now appears on the 2nd page. 

Screen Shot 2022-11-18 at 5 21 59 PMScreen Shot 2022-11-18 at 5 22 07 PM

With the pagination, should we still keep all assigned questions together on the first page? I also just added a 3rd section "All Assigned Questions" for professors in a separate PR. It's not merged yet, but that might also be something we need to account for with the pagination?

NIDHI2023:
Closing since not focusing development on this for maintenance mode

PR #693, opened 2022-11-20
TA Notifications

erin-xu:
### Summary 

This PR adds TA notifications and is a follow-up to my previous PR adding an "All Assigned Questions" section to the professor session view. Adding the "All Assigned Questions" sections allows professors to unassign TAs from questions. Currently, only the student is notified when this happens. This PR notifies TAs when they are unassigned from a student's question. The notification says: You've been unassigned from [student first name] [student last name]'s question.

### Test Plan 
Check to make sure that the TA is notified when a professor unassigns them from a question. Check that TAs do not receive a notification when they unassign themselves from a question. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -18,8 +18,8 @@ const client = new Twilio(accountSid, authToken);
 /**
  * Function that handles data and sends a text message to a requested phone number
  */
-async function sendSMS (user: FireUser, message: string) {
-    if(process.env.DATABASE === "staging") {
+async function sendSMS(user: FireUser, message: string) {
+    if (process.env.DATABASE === "staging") {
         return;
     }
     const userPhone = user.phoneNumber;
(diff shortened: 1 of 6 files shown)


functions/src/index.ts
```diff
@@ -405,20 +421,50 @@ exports.onQuestionUpdate = functions.firestore
...
                         notifications: admin.firestore.Timestamp.now(),
                         productUpdates: admin.firestore.Timestamp.now(),
-                        lastSent: admin.firestore.Timestamp.now(),})
-              
+                        lastSent: admin.firestore.Timestamp.now(),
+                    })
+
                 });
+                
+            if (answerer.userId !== sender) {
```
abagel21, line 441: I think you need to include the filtering from line 408--the update needs to be removing the answerer.userId, which means you should probably add "&& prevQuestion.answererId !== newQuestion.answererId && newQuestion.answererId === ''"to your existing condition.

abagel21 (changes requested):

NIDHI2023:
Closed since out of scope of maintenance mode

PR #695, opened 2022-12-04
Updates on Analytics Page

danakim21:
### Summary 

This PR makes changes to the existing Analytics page. The following changes are made: 

- There is a new section added which shows the average wait time per day 
- If there are multiple sessions a day, the data points are shown as 1 and the average wait time is the aggregate value 
- Date picker's undefined error is removed 
- The existing bar graph is removed 

** I need some feedback on the design since there is a big empty space from removing the initial bar graph. 

Screen Shot 2022-12-04 at 4 49 17 PM

### Test Plan 

- [ ] Check that the average wait time chart shows 
- [ ] Check that the average wait time is aggregated for multiple sessions a day 
- [ ] Change dates on the date picker and check that the undefined error does not appear 
- [ ] Check that the bar graph is removed 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AdminReadOnlyCourseCard.tsx
+++ b/src/components/includes/AdminReadOnlyCourseCard.tsx
@@ -24,7 +24,7 @@ const AdminReadOnlyCourseCard = ({ course }: { readonly course: FireCourse }) =>
                 <ul>
                     {course.professors.map(id => {
                         const professor = professorMap[id];
-                        if (professor == null) {
+                        if (professor === null) {
                             return null;
                         }
                         return (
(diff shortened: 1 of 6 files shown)


abagel21 (changes requested):
I tested that the average wait time graph showed, the bar graph didn't, that the data was correct and aggregated, and that no undefined error appeared with the date picker. The only issue is that when the date range has no resolved questions, the average wait time doesn't exist and thus becomes "NaN":
image

Please correct it to say something neater like "Not applicable" if there is no aggregate wait time.

abagel21 (approved):
Fixed lint and undefined. LGTM!

PR #696, opened 2022-12-05
Failed Demotion Message

erin-xu:
### Summary 

This is an extension to my previous PR preventing role demotions (professor to TA). This PR notifies the user of a failed role demotion. Previously, a message would appear to notify the user of any added roles or pending users. The message now also includes failed demotions. The number of lines in the message depends on how many categories apply (added users, pending users, failed demotions) as shown below. If none apply, the message does not appear. 

Screen Shot 2022-12-05 at 1 28 39 AM

Screen Shot 2022-12-05 at 1 29 03 AM

Screen Shot 2022-12-05 at 1 27 56 AM

This PR also resolves a bug with the ImportRolesModal. Before, when users exited out of the modal or submitted their role changes and opened the modal again, the previously entered role updates would remain. I updated the code to fix this. 

### Test Plan 
- [ ] Verify that the failed demotion message appears only when a demotion is entered. 
- [ ] Verify that after closing the import roles modal (clicking the x) or submitting role changes, the modal is empty when reopened (both manual import and csv upload options)

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (approved):
I tested that demotion prompts a message now, and that the modal correctly doesn't retain information after submission/closing. Thank you for fixing an additional bug! One small thing--I had to change line 155 from "data.length - 1" to "data.length" to get multi-line csv files to import correctly. I'm unsure how this hasn't been an issue before, so I'm curious if I was doing something wrong on my end or if that is a bug. I'm approving because the items in the scope of this PR are correct, but if that is a bug please rectify it in this PR.

PR #704, opened 2023-01-16
tweaking node memory variable

abagel21:
### Summary 

Build is crashing because there is not enough memory. Given that build succeeds locally, I believe this is simply because the github node variable for max memory is too low for our project now. I adjusted it in our workflows to attempt to fix it.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


triple-jay (approved):
Thank you for the fix Alex!

PR #718, opened 2023-02-12
Lily/addnetid

burninglilies:
### Summary 

Added net id to display with name in the topbar 

image

### Test Plan 

I ran this on my laptop as well as my monitor to make sure it displays correctly on both. And after resizing and testing display constraints I confirmed it displays correctly on different sizes.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ChatBox.tsx
+++ b/src/components/includes/ChatBox.tsx
@@ -1,0 +1,48 @@
+import React, { useEffect, useState } from 'react';
+
+import {connect} from 'react-redux'
+import addNotification from 'react-push-notification';
+import { Icon } from 'semantic-ui-react';
+import { logOut } from '../../firebasefunctions/user';
+import Logo from '../../media/QLogo2.svg';
+import CalendarHeader from './CalendarHeader';
+import ProfessorStudentToggle from './ProfessorStudentToggle';
+import TopBarNotifications from './TopBarNotifications'
+import {useNotificationTracker} from '../../firehooks';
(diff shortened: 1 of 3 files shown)


src/components/includes/ChatBox.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useEffect, useState } from 'react';
+
+import {connect} from 'react-redux'
+import addNotification from 'react-push-notification';
+import { Icon } from 'semantic-ui-react';
+import { logOut } from '../../firebasefunctions/user';
+import Logo from '../../media/QLogo2.svg';
```
Sydney-tran, line 7: I noticed that there are a lot of unused imports. Is there are reason for this or can they be deleted?

Sydney-tran (commented):
This looks good Lily! I just noticed that there are a lot of unused imports (and left a comment). I think only React, connect, and rootState are actually used so hopefully the rest of the imports can be cleaned up. But besides that, good job!

src/components/includes/ChatBox.tsx
burninglilies: Oh those were not supposed to be in my PR, I believe the reason for this was because I forgot to git pull before I pushed.

burninglilies: Wait never mind I did pull before I pushed

burninglilies: Ill slack you

PR #720, opened 2023-02-27
fixing undefined in admin

abagel21:
### Summary 

Not sure why this issue didn't rear its head previously but there were undefined errors in admin which caused it to not display. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AdminReadOnlyCourseCard.tsx
+++ b/src/components/includes/AdminReadOnlyCourseCard.tsx
@@ -29,9 +29,9 @@ const AdminReadOnlyCourseCard = ({ course }: { readonly course: FireCourse }) =>
                         }
                         return (
                             <li key={id}>
-                                {professor.firstName ?? professor.firstName}
-                                {professor.lastName ?? professor.lastName}
-                                ({professor.email ?? professor.email})
+                                {professor.firstName}
+                                {professor.lastName}
+                                ({professor.email})
                             </li>
                         );
                     })}
@@ -48,9 +48,9 @@ const AdminReadOnlyCourseCard = ({ course }: { readonly course: FireCourse }) =>
                         }
                         return (
                             <li key={id}>
-                                {ta.firstName ?? ta.firstName} 
-                                {ta.lastName ?? ta.lastName} 
-                                ({ta.email ?? ta.email})
+                                {ta.firstName} 
+                                {ta.lastName} 
+                                ({ta.email})
                             </li>
                         );
                     })}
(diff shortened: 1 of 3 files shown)


src/components/includes/AdminReadOnlyCourseCard.tsx
```diff
@@ -24,12 +24,14 @@ const AdminReadOnlyCourseCard = ({ course }: { readonly course: FireCourse }) =>
...
                     {course.professors.map(id => {
                         const professor = professorMap[id];
-                        if (professor === null) {
+                        if (professor === null || professor === undefined) {
                             return null;
                         }
                         return (
                             <li key={id}>
-                                {professor.firstName} {professor.lastName} ({professor.email})
+                                {professor.firstName ?? professor.firstName}
```
Sydney-tran, line 32: Since you already checked if the professor is null or undefined earlier do you still need to change this?

src/components/includes/AdminReadOnlyCourseCard.tsx
```diff
@@ -41,12 +43,14 @@ const AdminReadOnlyCourseCard = ({ course }: { readonly course: FireCourse }) =>
...
                     {course.tas.map(id => {
                         const ta = taMap[id];
-                        if (ta === null) {
+                        if (ta === null || ta === undefined) {
                             return null;
                         }
                         return (
                             <li key={id}>
-                                {ta.firstName} {ta.lastName} ({ta.email})
+                                {ta.firstName ?? ta.firstName}
```
Sydney-tran, line 51: Similar to professor, since you already checked if the ta is null or undefined earlier do you still need to change this?

Sydney-tran (commented):
This looks good! The only thing I am a little confused about is the use of the ?? operator. I also left comments about it, but from what I understand is if that the value on the left of the ?? operator is null or undefined, then the value on the right will be used as a fallback value. However, it seems that for each statement, the value on the left and right of the operator are the same so I am not too sure if this would truly catch a null or undefined value. Since you already made sure that the professor/ta was not null or undefined, I think you can assume that the firstName, lastName, and email should also not be null or undefined, but if this is not guaranteed maybe you could create and use default values for them. Besides that, good job! I am glad we can use admin page now!

src/components/includes/AdminReadOnlyCourseCard.tsx
abagel21: You are correct! the nullish operators are not necessary.

Sydney-tran (approved):
Looks good!

PR #722, opened 2023-03-01
Fix chat messages order

Sydney-tran:
### Summary 

1. Updates order of chat messages to be by ascending time rather than descending time
2. Moves reply/new comment to bottom of list of comments/chat messages

Before:
Screen Shot 2023-03-01 at 10 48 13 AM

After:
Screen Shot 2023-03-01 at 10 47 28 AM

### Test Plan 

Tested by adding, viewing, and deleting chat messages during office hours in both Student and Professor/TA view. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


erin-xu (approved):
This looks good, Sydney! I checked that the messages are displayed in ascending time and that new comments are added correctly to the bottom.

PR #724, opened 2023-03-05
Per TA Analytics

erin-xu:
### Summary 

This PR adds a graph to the professor analytics page to display the number of questions answered by each TA. Based on our discussion during standup, I moved the wait time graph to the top and added the TA graph to the bottom of the page as shown below. 
 
Screen Shot 2023-03-04 at 8 55 43 PM Screen Shot 2023-03-04 at 8 55 55 PM

Initially, the graph is empty since no TA has been selected.

Screen Shot 2023-03-04 at 8 56 07 PM

The professor can search for TAs by NetID. When the professor types in a sequence of characters, all students with NetIDs that begin with that sequence of characters will be displayed in a dropdown. The professor can scroll through and hover over names in the dropdown. 

Screen Shot 2023-03-04 at 8 56 33 PM

Once the professor clicks on a TA, the graph will display the questions answered by the TA during the date range selected. 

Screen Shot 2023-03-04 at 8 57 57 PM

Like the Most Crowded graph, different office hour session on the same day will be shown as separate sections of the same bar. 

Screen Shot 2023-03-04 at 9 08 44 PM

### Test Plan 
- [ ] Check that questions answered are displayed correctly for TAs across calendar date range and for days with multiple office hour sessions. When a different TA is selected, the graph data and TA info on the left should update. 
- [ ] Check that all TAs that have answered questions in the date range can be found through NetID. 
- [ ] Check that dropdown scrolls for a long list of TAs.
- [ ] Check that TA cell is blue when hovered over.
- [ ] Check that dropdown closes when a TA is selected or click outside.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/pages/ProfessorPeopleView.tsx
+++ b/src/components/pages/ProfessorPeopleView.tsx
@@ -46,11 +46,9 @@ const ProfessorPeopleView = (props: RouteComponentProps<{ courseId: string }>) =
     const [selectedTA, setSelectedTA] = useState<FireUser>()
 
     useEffect(() => {
-        const input = TAName.split(" ")
-        if (input.length !== 0) {
+        if (TAName.length !== 0) {
             const filtered = allTAs.filter((ta) =>
-                ta && ta.firstName.toLowerCase().startsWith(input[0]) &&
-                (input.length === 1 || ta.lastName.toLowerCase().startsWith(input[1])))
+                ta && ta.email.toLowerCase().startsWith(TAName))
             setFilteredTAs(filtered)
         } else {
             setFilteredTAs(allTAs)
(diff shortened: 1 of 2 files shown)


abagel21 (approved):
Tested different TAs and time ranges, looks excellent thank you! Just fix the lint errors before merging.

src/components/pages/ProfessorPeopleView.tsx
```diff
@@ -193,17 +243,26 @@ const ProfessorPeopleView = (props: RouteComponentProps<{ courseId: string }>) =
...
                                     </div>
                                 </div>
-                                {/* <div className="questions-bar-container">
-                                    <div className="bar-graph">
-                                        <QuestionsBarChart
-                                            barData={barGraphData}
-                                            sessionKeys={sessions.map((s) => s.sessionId)}
-                                            sessionDict={sessionDict}
-                                            yMax={chartYMax}
+                                <div className="Most-Crowded-Box">
```
Sydney-tran, line 246: Since there are so many nested components, a couple of your lines exceed the line length limit which is why there are lint errors. Maybe you could create a MostCrowdedBox component and then use it in ProfessorPeopleView.

Sydney-tran (commented):
This looks really good! I also left a comment about this, but one possible way to fix the lint errors about exceeding the line length limit could be to define some of the smaller components in separate files, and then use them in ProfessorPeopleView. This would help remove a lot of the nesting in ProfessorPeopleView. But besides that, very nice job Erin!

PR #727, opened 2023-03-15
Add current question time

Sydney-tran:
### Summary 

Displays the current minutes and seconds spent on a question to the assigned TA. Starts at the assigned time and ends time if unassigned, or marked as no show or done by TA.

Screen Shot 2023-03-15 at 4 55 13 PM

### Test Plan 

Tested by creating question in student view, and then assigning, unassigning, marking as no show, and marking as done in TA view.

### Breaking Changes 

If the question character length reaches the end of the question box, the question time will be pushed out of the box. However, most professors limit the question length, and most student's questions don't reach this length.

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/SessionQuestion.tsx
+++ b/src/components/includes/SessionQuestion.tsx
@@ -394,8 +394,7 @@ class SessionQuestion extends React.Component<Props, State> {
                     {(this.props.isTA || includeBookmark || this.props.includeRemove) &&
                         <p className={'Question' + studentCSS}>
                             {question.content}
-                            {this.props.isTA && question.status !== 'unresolved' && 
-                                question.timeAssigned !== undefined &&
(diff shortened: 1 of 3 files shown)


abagel21 (changes requested):
The time displays as expected. The only change is I think that the time should stop displaying if the office hour has ended because it isn't relevant at that point and could increment infinitely although it isn't particularly expected behavior. As far as it being pushed off-screen by text, I believe that is connected to the larger issue that unlimited text is incorrectly formatted in QMI in general, so I will fix that in a separate PR.

abagel21 (approved):
Looks great! I checked that it stops displaying when the office hour is past. I've also fixed (mostly) the issue of questions running over on another branch so that part should also be resolved.

PR #728, opened 2023-03-15
Add office hours calendar dots

Sydney-tran:
### Summary 

Adds gray dot below all days with office hours on calendar. Dot under day becomes invisible if that day is clicked, dot becomes visible again if a different day is clicked.

Screen Shot 2023-03-15 at 5 25 36 PM

Screen Shot 2023-03-15 at 5 25 57 PM

### Test Plan 

Made test office hours and tested clicking through days of the calendar.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CalendarDateItem.tsx
+++ b/src/components/includes/CalendarDateItem.tsx
@@ -7,8 +7,8 @@ type Props = {
     day: string;
     date: number;
     handleClick: Function;
-    numSessions: number;
-    activeSession: number;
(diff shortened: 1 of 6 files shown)


abagel21 (approved):
Tested switching between classes and days and confirmed the dots behave correctly re: our discussions at standup. I think that having the dots disappear makes the most sense behavior-wise. Nice work!

PR #730, opened 2023-03-22
Let admins change their own roles

abagel21:
### Summary 

Admin used to use (mostly) the default roles table. Updated it to allow mutating your own role if you are in admin view.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


burninglilies (approved):
Just looked at your commit and the change to make AdminView automatically skips checking for email verification makes sense!

PR #732, opened 2023-03-27
Add latest comment preview to session questions

Sydney-tran:
### Summary 

For every session question with comments/messages, if the reply chain is collapsed, then a preview of the latest comment is shown below the question. The preview is highlighted yellow for a student if a TA has added a message after they last opened the reply chain, and the preview is highlighted yellow for a TA if a student has added a message after they last opened the reply chain. This highlight goes away once the student/TA opens the reply chain/views the new message. There is also an expand button next to the preview that functions the same as the reply button.

Example of student view before opening reply chain:
Screen Shot 2023-03-27 at 11 04 40 AM
Example of student view after opening reply chain and closing again:
Screen Shot 2023-03-27 at 11 06 16 AM

Example of TA view before opening reply chain:
Screen Shot 2023-03-27 at 11 05 23 AM
Example of TA view after opening reply chain and closing again:
Screen Shot 2023-03-27 at 11 06 00 AM

### Test Plan 

- Created session questions and used both student and TA view to add and view messages with the reply chain open and closed
- Checked that the preview was highlighted only if the student/TA had not opened the reply chain since the TA/student had added a new message

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/LatestComment.tsx
+++ b/src/components/includes/LatestComment.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import Moment from 'react-moment';
+
+type Props = FireComment & {
+    poster: FireUser;
+    newComment: boolean | undefined;
+}
+
+const LatestComment = ({ content, timePosted, isTA, poster, newComment}: Props) => {
+
+    return (
+        <div className={"userComment " + (newComment ? "newComment Preview" : "Preview")}>
+            <div className="commentInnerWrapper">
+                {poster ?
+                    <div className="commentHeader Preview">
+                        <div className="commentNameAndProfile Preview">
+                            <img
+                                className="commenterProfile"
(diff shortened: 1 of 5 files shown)


erin-xu (approved):
Nice work, Sydney! I tested by adding multiple comments on both the TA and student sides with the comments both expanded and closed and verified that the latest comment preview is displayed and highlighted correctly. I checked that the expand button behaves correctly as well. For long comments, the preview correctly clips the content, but in the expanded chain, the content continues past the comment cell. I'm not sure if this is in the scope of your task, but it might be better if the text of longer comments is displayed on multiple lines?

Screen Shot 2023-04-16 at 10 53 33 AM

abagel21:
Also possible that my tutorial code PR (which includes fixes for comment styling on really long comments) would repair that if you're using the same styling.

PR #734, opened 2023-04-16
Mobile fixes

erin-xu:
### Summary 

This PR improves the mobile formatting of the top bar. 

Screen Shot 2023-04-16 at 10 26 53 AM Screen Shot 2023-04-16 at 10 24 58 AM

The screenshots above show a comparison between the old and new mobile views. On the new mobile view, the bug icon in the bottom left corner is hidden and the user name in the top right corner is hidden. Professor features are disabled as well: the Queue-Dashboard toggle is hidden and professors do not have access to the admin and product update pages from the mobile dropdown. The dropdown menus are also larger to improve usability. Let me know what you think about the adjusted dropdown menu size or if you have any other suggestions for changes to the mobile view!
Screen Shot 2023-04-16 at 10 13 21 AM Screen Shot 2023-04-16 at 10 13 05 AM

### Test Plan 
Switch to mobile view (adjust window width to be less than 920px).
- [ ] Check that the top bar does not extend outside the window.
- [ ] Check that the bug icon and user name are hidden.
- [ ] Check that the dropdown menus are larger.
- [ ] Login to an account with a professor role. Check that the Queue-Dashboard toggle is hidden on mobile view. Check that  the Product Update and Admin pages are hidden from the dropdown. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CalendarHeader.tsx
+++ b/src/components/includes/CalendarHeader.tsx
@@ -1,10 +1,10 @@
 import * as React from 'react';
 import { useHistory } from 'react-router';
-import {connect} from 'react-redux'
+import { connect } from 'react-redux'
 import { useMyCourses, useIsAdmin } from '../../firehooks';
 
 import { CURRENT_SEMESTER } from '../../constants';
-import {updateSession} from '../../redux/actions/course'
+import { updateSession } from '../../redux/actions/course'
 
 import Toggle from '../../media/Toggle.svg';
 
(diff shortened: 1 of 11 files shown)


src/components/includes/CalendarHeader.tsx
```diff
@@ -38,7 +38,15 @@ const CalendarHeader = ({ currentCourseCode, updateSession, role }: Props): Reac
         history.push(`/${page}`);
     };
 
+    const [width, setWidth] = React.useState<number>(window.innerWidth);
+    const MOBILE_BREAKPOINT = 920;
```
Sydney-tran, line 42: Not sure if it is that important, but maybe `MOBILE_BREAKPOINT` should be added to constants.ts rather than being redefined in every component file.

Sydney-tran (commented):
This looks really good Erin! I checked the mobile view and the top bar and dropdowns look much better now! The only thing I noticed is that the constant `MOBILE_BREAKPOINT` is used in many components, so rather than defining it over and over again, maybe the constant could be added to the constants.ts file. But besides that, very nice job!

burninglilies (commented):
I think it looks good! I ran it on my computer and it works as expected when I switch to mobile view and the professor related features are hidden.

NIDHI2023:
Added some of these changes to updated TopBar code. Mainly did not include changes to remove professor dashboard and removing bug icon on mobile, as these features are still usable on mobile.

PR #737, opened 2023-04-26
Filter questions by tags

Sydney-tran:
### Summary 

This PR adds filtering questions by tags for TA/professor view of session questions. First, the search and dropdown for filtering by category is shown, and then, once a category is selected, the search and dropdown for filtering by tags (which are the subtags of the selected category) can be shown by clicking on a button. Either filter can be cleared, where clearing a category filter will also clear the subtag filter.

https://user-images.githubusercontent.com/45403687/234701084-a85f2fac-062e-44aa-a783-ceab9938a590.mov

### Test Plan 

Tested by adding questions as a student and then filtering questions as a TA/professor. Checked filtering for both unanswered and answered questions by selecting, switching, and clearing different categories and different subtags.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


abagel21 (changes requested):
I tested the common workflow of adding a category and tag with multiple questions visible. It seemed to not work for unanswered questions:
image
For answered questions, it worked perfectly and the design was just as discussed at standup. My only note would be I think the distance between the category filter and subtag filter is too large when the latter is visible. I appreciate you working with ambiguity and producing a great feature! I think just fixing the unanswered questions and the spacing should do it.

burninglilies (approved):

PR #740, opened 2023-05-06
frontend implementation of course create

burninglilies:
### Summary 
This PR is front end implementation of course create design for professors and TA's to create new courses that will be pending and approved by a QMI user before being created as an official course.
image
image
image
image

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


Sydney-tran (commented):
Looks good so far! Just made a couple comments about deleting unused/commented out code and running yarn prettier to fix up some spacing issues. Besides the form titles (ex: Course Code) which need to be left aligned, the asterisks which need to be colored red, and the form examples which needed to be added in, everything looks very nice and matches the figma.

PR #751, opened 2023-09-18
Fixed Notifications Menu Bug

jewang25:
### Summary 

Fixed bug where the notifications menu would be hidden behind profile menu if profile menu was open before the notifications icon was clicked. Now, the menus switch if the different buttons are clicked in any order. Passed the showMenu variable and the function to change it down to the TopBarNotifications component.

### Test Plan 

Tested by running the website and clicking the notifications icon and the profile icon in different ways multiple times.

### Checklist

- [ x ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/TopBar.tsx
+++ b/src/components/includes/TopBar.tsx
@@ -99,7 +99,7 @@ const TopBar = (props: Props) => {
                         )}
                     </div>
                     <div className="rightContentWrapper" >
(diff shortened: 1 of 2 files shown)


burninglilies (approved):
Looks good! Also nice documentation 👍

PR #752, opened 2023-09-25
Update DTI branding on login page

rgu0114:
### Summary 

- Update DTI name to Digital Tech & Innovation on the login page
- Add updated icon

### Test Plan 

(look at the footer at the bottom)

268532959-dc9b876c-0156-4da6-9700-14bb76324bf2-1

### Breaking Changes 

None


Sydney-tran (commented):
Nice work Richard! PR looks good to merge. I just made one comment about whether or not we need/still use the old dti logo svg. If not, it should probably be deleted so we don't have unnecessary svgs. But we might need another PR to clean up the media folder in general.

burninglilies (approved):
Looks good! First released merge of semester :D
Also, yes, we can totally have cleaning up the media page be a potential PR in the future as well

PR #755, opened 2023-09-28
added sophie to readme for FA23

swang235:


b3njamint (changes requested):
Good job with your first PR, and thanks for doing it so fast! Could you add a 2023-2024 section and move your name under that instead (along with the rest of the team, you can find us in the QMI Slack channel).

PR #757, opened 2023-10-01
Autopopulate Test Data

erin-xu:
### Summary 
This PR adds buttons to the professor view page to add and delete test sessions. This feature will stay on the main branch and will be used by QMI members in the future for testing. The add button creates a session from 12:00am to 11:59pm for every day after the current day. The delete button deletes all sessions for the course. 

Screenshot 2023-10-01 at 2 37 18 PM

I had to make a few fixes in the process of implementing this:
- Update the current semester in `constants.ts` to FA23
- Fix the calendar week header. (The end date was one day later than it should be. So in the screenshot above, it used to be Oct 2 - Oct 9.)
- Update the start time range for retrieving sessions for a day. I changed the last line in the screenshot below from `<=` to `<`. Previously, sessions for 12:00am on October 9 would show up under October 2 because we were using `<=`.
    Screenshot 2023-10-01 at 2 47 05 PM

Let me know what you think about the placement of the buttons on the professor view page. For now, the add button creates one office hour session for each day, but would it be more helpful to populate the sessions a different way (multiple sessions per day, not every day, etc)? 

### Test Plan 
- [ ] Verify that the dates in the calendar week header are correct. 
- [ ] Check that clicking the `Auto Populate Sessions` button create one office hour session for the current day and every day that follows.
- [ ] Check that clicking the `Delete All Sessions` button deletes all office hours sessions for the course.
- [ ] Verify that office hour sessions appear under the correct dates in the professor calendar view.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/CalendarWeekSelect.tsx
+++ b/src/components/includes/CalendarWeekSelect.tsx
@@ -79,7 +79,7 @@ class CalendarWeekSelect extends React.Component<Props, State> {
 
         const thisWeekEpoch = this.state.selectedWeekEpoch;
         const nextWeekEpoch = this.state.selectedWeekEpoch +
-            7 /* days */ * 24 /* hours */ * 60 /* minutes */ * 60 /* seconds */ * 1000 /* millis */;
+            6 /* days */ * 24 /* hours */ * 60 /* minutes */ * 60 /* seconds */ * 1000 /* millis */;
 
         return (
             <div className="CalendarWeekSelect">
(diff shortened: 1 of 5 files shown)


swang235:
Hi Erin, nice work! I like the placement of the buttons, I think the order makes sense. From my understanding, the purpose of the new buttons is just to make testing easier? I'm not too familiar with the current tasks that would benefit from the autopopulation, but I think making 1 daily OH session is fine!

burninglilies (approved):
Looks good! As a next step we can work on auto populating students in a queue somewhere only admins can access after we have a design so that we can use this to identify bugs in the future such as students disappearing from queue

PR #762, opened 2023-10-04
Format links and linkify comments

burninglilies:
### Summary 

Changed display of Questions from flex since Linkify makes each link a new component. Change in formatting shouldn't really affect comments in the future because comments have a character limit anyways. 
Linkified comments for both students and TA's 

TA View:
image

Student View:
image

Latest comment preview is intentionally not linkified:
image

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/styles/session/SessionQuestionsComponent.scss
+++ b/src/styles/session/SessionQuestionsComponent.scss
@@ -225,7 +225,7 @@
     margin: 0em;
     color: #595959;
     font-weight: 300;
-    display: flex;
+    // display: flex;
     align-items: center;
     justify-content: space-between;
   }
--- a/src/styles/session/SessionQuestionsComponent.scss
+++ b/src/styles/session/SessionQuestionsComponent.scss
@@ -225,7 +225,6 @@
     margin: 0em;
     color: #595959;
     font-weight: 300;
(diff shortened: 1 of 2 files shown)


src/styles/session/SessionQuestionsComponent.scss
```diff
@@ -225,7 +225,7 @@
     margin: 0em;
     color: #595959;
     font-weight: 300;
-    display: flex;
+    // display: flex;
```
Sydney-tran, line 228:
```suggestion
```

Sydney-tran (commented):
Nice work Lily! Linkify-ing the comments will definitely be helpful for communication between students and TAs. One small comment I have is that you commented out some styling, but if you don't need it you can probably just remove that line to keep the files a little cleaner. Besides that, the PR looks good!

Just a related thought that I can also bring up at standup: the blue color that Linkify uses to signify that there is a link is quite similar to the QMI blue we use for text in some other places on the website that aren't links. This might be confusing for users so we might want to update it.

rgu0114 (approved):
Agree with Sydney's comments, also looked good during work session demo. LGTM!

PR #763, opened 2023-10-12
Add pending courses in prof view 

Sydney-tran:
### Summary 

This PR adds the backend for creating a new pending course as a professor/ta. On submit, a course is created from the information in the form and added to the pendingCourses collection. If the courseId (course code, year, term) already exists in the pendingCourses or courses collections, then an error is thrown.

Note: Frontend for prof view and backend for admin view is done in another PR

Future todos: 
- Currently, only adds to pending courses if the inputted semester (year and term) matches the current semester => need to finish implementation if inputted semester does NOT match current semester
- Currently, the error message is only logged in console => need to design and implement showing the user the error message (for when courseId already exists)

Documentation:

### Test Plan 

- [ ] Test frontend and check form works as intended
- [ ] Test backend and check course is added to the pendingCourse collection as intended
- [ ] Test edge cases such as trying to create course that already exists

### Notes 

- Firebase must be updated to include pendingCourses
- PR must be merged in at same time as PR for backend of admin side of pending courses

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rgu0114 (commented):
This looks great, nice work! I like how extensively you tested the Firebase functions and also documented what future work needs to be done as the next step to implementing this feature. I was wondering though, how would these pending courses be approved? Is there a task out for how that would extend from your work in this PR?

PR #764, opened 2023-10-22
Richardgu/edit roles bug

rgu0114:
### Summary 

This PR addresses a bug with the edit role dropdown for the TA role selected, where much of the component is cut off on the right side.

image

To address this, we change the Dropdown component to open to the left instead of to the right.

### Test Plan 

The Dropdown is no longer getting cut off after this change

Screen Shot 2023-10-22 at 12 42 02 AM

### Notes 

Note that for the other roles (student and professor), the Dropdown appears slightly to the left from where it was previously 

Screen Shot 2023-10-22 at 12 42 33 AM

Screen Shot 2023-10-22 at 12 43 01 AM

### Documentation 

[Note on the rationale for the exact fix](https://github.com/cornell-dti/office-hours/pull/764/files)

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorRolesTable.tsx
+++ b/src/components/includes/ProfessorRolesTable.tsx
@@ -33,6 +33,10 @@ const RoleDropdown = ({
                         changeRole(user, course, newValueRole);
                     }
                 }}
+                /**
+                 * Dropdown alignment left ensures the component expands to the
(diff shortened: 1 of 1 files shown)


erin-xu (approved):
Nice work, Richard! I verified that the dropdown is no longer cut off.

PR #765, opened 2023-10-23
Per Student Analytics [WIP]

erin-xu:
### Summary 

This PR adds a  graph to the professor analytics page to display the number of questions asked by each student. I added it below the per TA analytics graph and used the same format as well. 

Screenshot 2023-10-22 at 7 59 25 PM

Professors can search for a student using net ID and hover over the graph to view the office hour location, time, and number of questions asked. If a student asks questions during multiple sessions on the same day, the corresponding bars are stacked on top of each other in the graph.

Screenshot 2023-10-22 at 8 05 48 PM

** The functionality is done but I marked this PR as WIP because I want to go back to clean up my code and create a new component for the per TA/student graph so there is less code duplication.

### Test Plan 
- [ ] Check that questions asked are displayed correctly for students across calendar date range and for days with multiple office hour sessions. 
- [ ] When a different student is selected, the graph data and student info on the left should update.
- [ ] Check that all students can be found through NetID.
- [ ] Check that dropdown scrolls for a long list of students.
- [ ] Check that student cell is blue when hovered over. 
- [ ] Check that dropdown closes when a student is selected or click outside.
- [ ] Check that the y-max of the chart is correct - 8 by default and the max number of questions answered in one day if it exceeds 8.

### Breaking Changes 

None

- [ ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore

PR #767, opened 2023-10-31
Full-product Analytics Page

b3njamint:
### Summary 

This PR is the first step towards implementing an analytics page for Queue Me In as a whole, rather than just for individual courses.

- Queries aggregate counts (courses, questions, users, sessions) from Firebase collections for the entire history of QMI
- Takes current time to automatically query data from the current semester
- Note: Reads the entire collection all together as one so it shouldn't drastically increase read count (I think)

It is just a page for internal use so it looks kinda bad——might add styling in the future.

**Update**: Added some styling:
Screen Shot 2023-10-31 at 2 17 16 AM

### Test Plan 

I added things in the test Firebase and the numbers changed.

Snapshot of the analytics page.
Screen Shot 2023-10-31 at 12 41 42 AM

Snapshot of the analytics page after I added another FA 23 course. Observe that both the historical and current course count increments, as we are currently in FA 23.
Screen Shot 2023-10-31 at 12 44 05 AM

### Breaking Changes 

Some library versions changed but it shouldn't affect anything.

- [ ] My PR adds a @ts-ignore


Sydney-tran (commented):
Nice work Benji! The code looks good and is commented very nicely. I think having an analytics page for qmi as a whole will definitely be useful for internal use and using a table to present the data definitely works well. One possible next step could be turning the [per student analytics](https://github.com/cornell-dti/office-hours/pull/765) for each course into overall qmi student analytics.

rgu0114 (changes requested):
Thanks for taking the time to work on this page Benji! It'll definitely provide a lot of useful insights for the product as a whole and make it easier to identify areas where we can improve. Since this does appear to be for internal use, I'm ok with pushing this out as a first iteration and then continuing to add to it next semester. Or we can keep building/iterating on this until it's fully ready to be released to production, but I think this provides a lot of utility that could be useful in the meantime.

On a first read through, the logic of the analytics page seems to be correct and match with its intended purpose. I'll review the code more closely in the future, but for now I don't think it's that intuitive that we'd need to navigate to `/analytics` to see this product analytics page. It might be useful to have some way to easily get this page from `/admin`, especially since I'd imagine the users of the admin page and this new analytics page would mostly be the same.

PR #775, opened 2023-11-12
Rename tags

erin-xu:
### Summary 
This PR adds functionality to rename tags for an assignment in the `Manage Tags` portion of the professor dashboard. In addition to adding new tags and deleting existing tags, professors can now click into existing tags and change the tag name. 

Screenshot 2023-11-12 at 6 18 34 PM

### Test Plan 
- [ ] Verify that tags can be edited by clicking into the tag and typing in a new name.
- [ ] Verify that when students select tags for their question, the updated tag names are displayed. 
- [ ] The tag border should turn blue when hovering over the tag. 
- [ ] The X should turn blue when hovered over. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorTagInfo.tsx
+++ b/src/components/includes/ProfessorTagInfo.tsx
@@ -89,6 +89,15 @@ class ProfessorTagInfo extends React.Component<PropTypes, State> {
         }));
     };
 
+    handleModifyTag = (oldTag: string, newName: string): void => {
+        this.setState(prevState => ({
+            newTags: prevState.newTags.map(
+                (tag) => tag.name === oldTag ? { id: tag.id, name: newName } : tag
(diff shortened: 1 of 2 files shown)


jewang25 (commented):
Nice job Erin! I tested it out and it works well :). I like how you built off the existing code and added a new function to handle modifying tags. I think you can create multiple tags with the same name which doesn't really make sense, but not sure how relevant that is to your task haha.

burninglilies (approved):
Great work! I like the idea of having an input element to make it tags have editable text box so that you can edit all tags at once by clicking in instead of enabling and disabling an edit mode for each edit tag yourself. Also having changed color when hovered was a nice touch as well.

PR #779, opened 2023-11-29
Collapse former classes

erin-xu:
### Summary 

This PR adds functionality to collapse and expand former courses on the classes page. When the page is initially loaded, the former courses default to collapsed. 

Screenshot 2023-11-28 at 6 23 13 PM

Screenshot 2023-11-28 at 6 25 47 PM

### Test Plan 
- [ ] Check that the former courses can be collapsed and expanded by toggling the icon
- [ ] Check that the former courses default to collapsed 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rgu0114 (approved):
approving

PR #781, opened 2023-12-03
Fix Notifications Styling

jewang25:
### Summary 

- fixed notifications dropdown styling so that it matched the user profile dropdown (picture below) more closely 
- mainly just changed font and the box css
 Screenshot 2023-12-02 at 8 52 16 PM
- may need to double check with design to make sure it is ok 

### Test Plan 

- tested by creating different notifications and seeing if it looked ok on the frontend (see some scenarios below)
Screenshot 2023-12-02 at 9 01 41 PM
Screenshot 2023-12-02 at 8 52 08 PM
Screenshot 2023-12-02 at 9 03 46 PM
Screenshot 2023-12-02 at 9 04 03 PM

### Notes 

- may need to reconsider notifications design in the future if really want to align the designs because it is pretty different from the user profile dropdown 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/TopBarNotifications.tsx
+++ b/src/components/includes/TopBarNotifications.tsx
@@ -31,6 +31,10 @@ const TopBarNotifications = ({notificationTracker, user, showMenu, iconClick}: P
 
     const dropdownRef = useRef<HTMLDivElement>(null);
 
+    /**
+     * This function calls a firebase function that keeps track of which 
+     * notifications have been viewed and updates the firebase 
+     */
     const updateTrackable = () => {
         viewedTrackable(user, notificationTracker, true)
     }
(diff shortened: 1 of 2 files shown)


swang235:
hi jessica, the notifications work properly for me! I think it would be nice for the user to be able to click on the notifications (ex its not blue anymore and theres no red circle on the notif symbol if they've viewed it already)- im not sure if that was included in the designs, but otherwise nice work!

rgu0114 (approved):
Hey Jessica, I agree with Sophie here. The styling and alignment look great! I do see that the dropdown is see through now though – just checking if that's expected?

PR #790, opened 2024-02-18
Made logo clickable to go to the home page

stevenyuser:
### Summary 

Made the QMI logo clickable, so that when it is clicked, it redirects the user to the "/home" page.

image

image

### Test Plan 

Tested clicking the logo on different screen sizes (mobile, computer, tablet).

### Notes 

Made the logo have an underline on hover. Also changed the user profile underline hover behavior to be consistent in style with the logo hover style.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/styles/TopBar.scss
+++ b/src/styles/TopBar.scss
@@ -57,14 +57,6 @@
             justify-content: center;
         }
 
-        // .logo {
-        //     align-self: right;
-        //     transition-duration: 0.2s;
-        //     transition-property: box-shadow;
-        //     transition-timing-function: $curve-standard;
-        //     cursor: pointer;
-        // }
-
         .QMILogoImage {
             align-self: right;
             padding-bottom: 10px;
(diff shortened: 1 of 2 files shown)


src/styles/TopBar.scss
```diff
@@ -57,12 +57,29 @@
             justify-content: center;
         }
 
-        .QMILogo {
+        // .logo {
```
rgu0114, line 60: make sure to remove all comments before pushing

rgu0114 (changes requested):
This looks great Steven! Thanks again for taking this on and for making sure to test on tablet, mobile, etc. We may want to check with design/product on the color of the underline, and make sure to remove commented out code. I'll approve as soon as those two things are resolved.

stevenyuser:
Thanks for the feedback, Richard! I just removed the commented-out code.

burninglilies (approved):
Thanks for the changes Steven, I double checked the hover coloring for the logo shadow and it's consistent with the user profile hover. So looks good!

rgu0114 (approved):

PR #791, opened 2024-02-22
Added .nvmrc to set correct node version

stevenyuser:
### Summary 

Added a .nvmrc file, which automatically sets the correct node version (v14.17.3) for development after running the command `nvm use`. (Useful for devs who have multiple node versions on their computer).

### Test Plan 

N/A

### Notes 
To set the node version to the correct version as specified in the .nvmrc file, you must run `nvm use` each time you open up your terminal or configure your terminal to automatically run `nvm use` when it detects a directory has a .nvmrc file.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rgu0114 (approved):
Thanks for thinking of doing this, Steven. Definitely appropriate for qmi and will be useful for many of us that keep having to switch between different node versions. Going to approve this, feel free to merge during work session today.

PR #797, opened 2024-02-22
Leave queue popup v2

rgu0114:
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


burninglilies (approved):
Tested it just now it it looks good! No node sass error and everything works as expected. The leave queue body/prompt/container css styling is a good workaround to avoid inline styling. I also liked that you added mobile formatting, which also looked good when I texted it as the popup was able to maintain shape!

stevenyuser (commented):
Looks good, Richard! I tested it for desktop and mobile screen sizes and the styling is good on both of them. Good responsive design! Also, I liked your documentation in LeaveQueue.tsx and how you explained the code.

PR #805, opened 2024-03-08
Collapsible TA/Professor view

NIDHI2023:
### Summary

- added icons to collapse professor and TA lists

- editing and roles button now change color when the cursor hovers over them

Screen Shot 2024-03-07 at 9 37 53 PM

image

### Notes

- icons should appear when there is at least one person in the TA/Professor list, otherwise only "None" should be displayed

### Test Plan 

- Hover mouse over editing and roles button to see if they change
- Check that icons default to being collapsed
- Check that clicking icons toggles between collapsed and expanded list
- Change users to be different roles and see if the icons appear accordingly

### Checklist

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/AdminReadOnlyCourseCard.tsx
+++ b/src/components/includes/AdminReadOnlyCourseCard.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useCourseProfessorMap, useCourseTAMap } from '../../firehooks';
+import { Icon } from 'semantic-ui-react';
 
 const AdminReadOnlyCourseCard = ({ course }: { readonly course: FireCourse }) => {
     const professorMap = useCourseProfessorMap(course);
     const taMap = useCourseTAMap(course);
+    const [collapsed, setCollapsed] = useState(true);
 
     return (
         <div>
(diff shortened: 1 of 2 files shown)


rgu0114 (changes requested):
Great work on this Nidhi! This is exactly what we were looking for, and it will help us navigate the admin page much more easily in the future. I also like how you added the hover effect to the edit mode/show roles buttons. 

One small change though that I know I didn't mention when we first assigned you this task but would improve how it looks. Could you show the first two professors and the first four TAs by default before the icons appear? Then have the icon appear below that list to expand it further. Now that I'm seeing a v1 for this, I think it looks a little off if we don't show any profs/TAs at all. 

Other than that, really good work and nice job making it through this fairly quickly!

burninglilies:
Great PR so far! In terms of functionality your code is good, the interactivity of hover over button feature was a great touch! I just have some comments about aesthetics, I agree with Richard that collapsing to a minimum of 4 TA's/Professors might be a bit better for aesthetic purpose, and the color of the hover button looks a little off to me since it doesn't go with the color scheme of the admin page. 

One small suggestion I have is looking into tertiary operators for conditional rendering so that you wouldn't have to use so many if statements. For example

Instead of 
`if (list.length <= defaultSize) {
 return <your code>
}
return <your code>
`

You could do
'(list.length <= defaultSize) ?
<your code if true> :
<your code if false>

rgu0114 (commented):
Thanks for making the additional changes Nidhi! Functionality works great, nice job figuring out how to do the initial state. Small nit: I think it might look better if the dropdown arrow was under the rest of the prof/TA names when we click expand so it looks like more of a natural progression. Other than that and what Lily suggested, I think this PR is looking pretty close to finished.

Screen Shot 2024-03-25 at 7 42 01 PM

rgu0114 (approved):
force push 😦 changes look good though, approving now and we can merge during work session

PR #810, opened 2024-03-21
Upgrade Node to v16.20.2 and other dependencies

rgu0114:
### Summary 

- Made the long needed upgrade from v14.17.3 -> v16.20.2
- Future upgrades to come, but this was the minimum requirement for some necessary packages
- Upgraded TypeScript to latest to make it compatible with node version
- Yarn upgraded all dependencies for good measure

### Test Plan 

- normal QMI functions behave as expected

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


burninglilies (approved):
Great PR! I agree with upgrading to node dart since sass depreciated. Behavior of QMI also works as normal on my end and there appears to be no console errors.

PR #811, opened 2024-03-21
Bug fix: Enabled deleting OH

burninglilies:
### Summary 

Previously we were not able to delete an OH session/series if it had passed. I removed the logic for checking if an OH has passed so that all OH sessions would be deletable.

### Test Plan 

Before: 
image
After:
image

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


NIDHI2023 (commented):
The deletion works for me for office hours that are currently in session and future office hours. I think it would be helpful if the buttons didn't look greyed out though, so it's clear that you have the option to delete.

rgu0114 (approved):
Agree with Nidhi, appears to work as intended. This fix will be helpful in producing less questionable and confusing behavior.

PR #814, opened 2024-03-24
Update README with SP24 Designers 

rgu0114:
### Summary 

Added Lucy and Sia to the list of contributors on the README

### Test Plan 

QMI still works

### Breaking Changes 

None


NIDHI2023 (approved):
Added in the right place, looks good!

PR #818, opened 2024-03-28
Fix Bug That Prevented Editing OH Time

stevenyuser:
### Summary 

- Fixed bug that prevented CS 2110 Course Staff from editing start time for a hybrid weekly recurring office hours
- Description of bug:
  - Essentially, in each of CS 2110's session documents in Firestore, the TALink field was missing
  - When TALink is pulled from the CS 2110 session document, it is undefined
  - This undefined TALink is passed into updateSession (which calls firestore.collection().doc().update()) when one tries to edit OH time
  - This returns an error because updating a document with a field with a value of undefined is unsupported by Firebase
- Fix:
  - When TALink is undefined, made the default value of TALink an empty string ("")
  - This prevents TALink from being undefined, which allows updateSession / firestore.collection().doc().update() to be called without errors

### Test Plan 
- Check that editing office hours session times works

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/ProfessorOHInfo.tsx
+++ b/src/components/includes/ProfessorOHInfo.tsx
@@ -55,8 +55,9 @@ const ProfessorOHInfo = (props: {
     );
     const [useTALink, setUseTALink] = useState(session && 
         (session.modality === "virtual" || session.modality === "hybrid") ? session.useTALink :false);
-    const [TALink, setTALink] = useState(session && 
-        (session.modality === "virtual" || session.modality === "hybrid") ? session.TALink : "");
+    const [TALink, setTALink] = useState(
+        session && (session.modality === "virtual" || session.modality === "hybrid") ? session.TALink || "" : ""
+    );
 
     React.useEffect(() => {
         if (session) {


NIDHI2023 (changes requested):
The editing for an existing office hour looks like it's working good! I don't seem to be able to directly add a weekly recurring office hour as a professor (doesn't show up) so I can't actually test editing weekly recurring hybrid office hours though.

Some additional things I saw while testing: 

- In a professor role, when this button is checked for a non-weekly recurring office hours the deletion doesn't seem to work. It works without the checkmark however, and I'm not sure if this has the same behavior for weekly recurring office hours.

image

- If you try to create an in person office hour that's weekly recurring without a location, you are prompted to put one. If you try to do this for a not weekly recurring office hour, it lets you. Not sure if that was intended or not?

rgu0114:
> The editing for an existing office hour looks like it's working good! I don't seem to be able to directly add a weekly recurring office hour as a professor (doesn't show up) so I can't actually test editing weekly recurring hybrid office hours though.
> 
> Some additional things I saw while testing:
> 
> * In a professor role, when this button is checked for a non-weekly recurring office hours the deletion doesn't seem to work. It works without the checkmark however, and I'm not sure if this has the same behavior for weekly recurring office hours.
> 
> image
> 
> * If you try to create an in person office hour that's weekly recurring without a location, you are prompted to put one. If you try to do this for a not weekly recurring office hour, it lets you. Not sure if that was intended or not?

@NIDHI2023 It was good that you noticed these other inconsistencies/issues with editing office hours, thanks for doing such thorough testing. To provide a bit of context, those are other things we plan on addressing on dev (sometimes not even a problem on prod) in the future, but unrelated to this PR. The recurring OH deletion especially is a bug we've tried to figure out for a while but haven't found a solution yet.

rgu0114 (approved):
Thanks for providing a simple fix to this bug and providing great detail for the rationale behind it! This will definitely help us solve a pretty pressing issue for 2110, so good work helping us get this out in time for deployment later this week.

PR #819, opened 2024-03-28
Upgrade TypeScript to v4.1.6 

rgu0114:
### Summary 
Upgraded TypeScript to v4.1.6

Some notes on changes:
Added the 
`"**/@typescript-eslint/eslint-plugin": "^4.1.1",`
    `"**/@typescript-eslint/parser": "^4.1.1"`
lines to resolve the `0:0  error  Parsing error: Cannot read property 'map' of undefined` linting pipeline error we see from upgrading TS

That brings up a whole host of new (but more fixable) errors. From there, it's a matter of going file by file to ensure everything is strongly typed and using the newer anonymous function type annotations, rather than the older `Function` annotation.

### Test Plan 

No more errors, even momentary ones that disappear, when running locally. Normal QMI functions work as intended.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,9 @@
   },
   "resolutions": {
     "@types/react": "^16.8.4",
-    "@types/react-dom": "^16.8.4"
+    "@types/react-dom": "^16.8.4",
+    "**/@typescript-eslint/eslint-plugin": "^4.1.1",
+    "**/@typescript-eslint/parser": "^4.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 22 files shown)


burninglilies (approved):
Upgrading TypeScript requires updating all our previous functions that used depreciated methods. Thanks for going through and updating the functions as a part of this process! Branch changes have no errors and QMI behavior still works as intended when I tested. Overall, looks good

PR #823, opened 2024-03-30
[Draft] prof ta invite email

rgu0114:
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


stevenyuser (commented):
The firebase cloud function onPendingUserCreate looks good to me. 

A possible improvement is to put more information in the email (i.e. what course the pending user was invited/imported in), so that the professor knows which course they've been invited to sign up for. 

We already have some of this information in the documents of "pendingUsers/{userEmail}" as under the roles property, we store `[course.courseId]: role` ([as shown here](https://github.com/cornell-dti/office-hours/blob/26bf604856341d4fe762a0d32fac908d78bbc9f7/src/firebasefunctions/importProfessorsOrTAs.ts#L50)). However, I think we would need to create another field with course name, which won't be hard, as we already pass a FireCourse into the function. 

Afterwards, we could add some code in the text of the email to something like: "You've been invited to be a [role] in [course name] on QueueMeIn! Click here to sign up: https://queueme.in"

NIDHI2023 (approved):
I agree with Steven's comment, adding the role and course name would improve the email message and make it more specific. I also think it's definitely better to use a .env file instead of functions.config to adhere to code standards.

PR #825, opened 2024-04-10
Filter admin courses

NIDHI2023:
### Summary 
- Adds a selection option for courses by semester to admin page
- Defaults to the current semester in the codebase 
  - seems to be SP23, not sure if that's supposed to be updated?

image

- Displays Archived Courses section only for default semester, only displays filtered courses for rest of the semesters
- Changed "create course" button theme to match QMI theme more

image

Dependency Changes:
- run yarn add @mui/material @emotion/react @emotion/styled

### Test Plan 

-Check if all the desired semesters are available in the dropdown
-Check if the admin page loads the expected default semester, and the archived courses are only visible for this semester
-Check if the correct courses are displayed for each semester
-Check if the "Other" category displays any extraneous courses
-Check if creating a new course while the courses are in their filtered display properly shows up

### Checklist

- [ x] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore


rgu0114 (commented):
I'll also add more code comments sometime tomorrow + we'll loop in the designers to see if they have any suggestions.

rgu0114:
Could you try making the filter selection background white so that it doesn't blend in with the grayish page base color too much? There should be an option to do that. A bit wider by default as well.

Screen Shot 2024-04-15 at 6 47 15 PM

rgu0114 (approved):
lgtm!

PR #826, opened 2024-04-14
Added more detailed time statistics to the Question Analytics page

stevenyuser:
### Summary 
Added more detailed wait time statistics to the Question Analytics page
- On the bar graph, include the average wait time for each session
- Refined the average wait time into a formatted string
- Added an average resolve time section to let professors see the average amount of time TAs spend answering students' questions

image
image

### Test Plan 

- Check that average wait time and average resolve time are calculated correctly (compare calculations to other parts, like SessionInformationHeader.tsx)

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/components/pages/ProfessorPeopleView.tsx
```diff
@@ -84,15 +84,34 @@ const ProfessorPeopleView = (props: RouteComponentProps<{ courseId: string }>) =
...
+        if (timeHours === 0) {
+            return timeDispMins + " mins " + timeDispSecs + " s";
+        }
+        return timeHours + " h " + timeDispMins + " mins";
+    };
+
+    // average wait time in seconds
     const calculateAverageWaitTime = (session: FireSession) => {
-        return session.assignedQuestions === 0 ? 0 : session.totalWaitTime / session.assignedQuestions / 60;
+        return session.assignedQuestions === 0 ? 0 : session.totalWaitTime / session.assignedQuestions;
```
rgu0114, line 107: Forgot if we discussed this, why are we removing the divide by 60 here for seconds instead of minutes?

rgu0114 (changes requested):
Thanks for taking this task Steven, overall looks correct to me and I agree with the logic of what you did. I think it will also be helpful for professors to be able to see the avg wait types in the ways you provided. I think after Lily takes a look at this as well this should be safe to merge.

In general though, for analytics related PRs it's helpful to provide a bit more description and justification for your changes, as well as showing testing to make sure that the wait times are correct. I think this would also be a good place to add documentation, particularly for the denser functions you write, for this upcoming dev portfolio.

rgu0114 (changes requested):
Also remember to fix the linting issues. Appears to just be complaining abt indentation so should be quick.

src/components/pages/ProfessorPeopleView.tsx
rgu0114: oh nvm i see

PR #854, opened 2024-04-25
QMI Wrapped Frontend

rgu0114:
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


NIDHI2023 (changes requested):
I was looking online and it seems to say you should avoid using React.FC to type your components because it can implicitly take children even if you don't want it to. Maybe you could avoid using the 'React.FC<.....>' pattern by rewriting it with explicit props and an implicit (or explicit) return type. So for example with this function:

> const Dot: React.FC<DotProps> = ({ active }) => (
    <div className={`dot ${active ? 'active' : ''}`} />
);

you could rewrite as:

const Dot = ({ active }: DotProps) => (
    <div className={`dot ${active ? 'active' : ''}`} />
);

This is the style I see in a lot of other functions in the codebase too.

Also since we're using a lot of unique media for QMI Wrapped specifically, and I don't think it'd be used very often elsewhere, I think it'd help organization to make a 'Wrapped' folder in /media. Overall though, this is a great base to work off of and it's very easy to understand!

NIDHI2023 (approved):

PR #862, opened 2024-05-04
Exporting calendar redesign

NIDHI2023:
### Summary

Implements Lucy's design changes for the exporting Google/Apple calendar feature.

Originals:

image
image

Figma redesigns:

image
image
image

### Actual Changes

- Got rid of bookmark
- Changed TA unfinished indicator to be a red dot when there is someone in "Ahead"

image

image

Redesigned modal:

- Buttons are now vertical
- Functionality is the same
- Got rid of repeating office hour title, now is just course code and OH name
- Added date and time as subtitle

image

With overflowing text:

image

### Test Plan

- Change the "Ahead" number of students and make sure red dot appears when there is at least one, otherwise should go away
- Make sure Calendar Export modal shows expected name, date, and times for any office hour
- Check if styling remains consistent for long OH names
- Check if buttons work as expected

### Notes

- Added dependency (ran yarn add @mui/icons-material)

### Breaking Changes 

None

### Checklist

- [ ] My PR adds a @ts-ignore


stevenyuser (commented):
Hi Nidhi, the redesign looks pretty good! I think the design very closely matches Lucy's Figma designs and looks a lot better than the previous version.

I have some minor suggestions/bugs:
- There's a CSS error when the location of the OH session is two lines
image

- When clicking "Add to GCal", I think it should open a GCal link on a new tab instead of opening the GCal link on the same tab. One way to do this is by changing your current GCal Button from this:
```js
<Button href={getGoogleCalendarLink()} className="export-btn" >
    <Button.Content icon>
        <Icon name='google' />
        Add to GCal
    </Button.Content>
</Button>
```
to this:
```js
<Button
    onClick={() => window.open(getGoogleCalendarLink(), "_blank")}
    className="export-btn"
>
    <Button.Content icon>
        <Icon name="google" />
        Add to GCal
    </Button.Content>
</Button>
```

Other than that, everything looks great! Good job!

burninglilies (commented):
Hi Nidhi! Great work on this speedy PR. Aside from Stevens suggestion about opening Calendar in new tab and trying to make Session names one line I just have one more suggestion. Since we are no longer using the bookmark icon nor Google or Apple Icon it would also be a good to remove them from our media as well if they are not being used.

NIDHI2023:
- Fixed Steven's suggestions for styling issues and opening a new tab

image

- Added a small fade in animation for the calendar modal only (did not touch the others in case this animation is not needed, but can be easily applied to modals in general too)

-Unneeded icons were deleted from /media

src/components/includes/CalendarExportModal.tsx
```diff
@@ -142,43 +149,51 @@ const CalendarExportModal = ({
...
+                                fontSize="large" 
+                                sx={{
+                                    color: "black"
+                                }} 
+                            />
                         </button>
-                        {isDayExport ? 
-                            <div className='Title'>{getDateString()} Office Hours</div> :
+                        <img src={CalIcon} alt='Calendar export icon' id="calIcon" />
+                        {isDayExport ?
```
rgu0114, line 160: nit: for code readability, make it more clear what's being conditionally rendered. For longer components like these, it might be helpful to extract them into helper functions. So in this case I might do the following instead...

```
const DayExport = () => { <div> ... </div> }

const SessionExport = () => { <div> ... </div> }

{isDayExport ? `<DayExport />` : `<SessionExport />`}


```

rgu0114: also there's quite a bit of repeated code here – you can try destructuring `currentExportSessions` earlier in the code or moving `<div className='Title'...` outside the ternary, for example. Play around with these options and see which combination looks cleanest to you!

src/components/includes/CalendarSessionCard.tsx
```diff
@@ -100,12 +93,16 @@ const CalendarSessionCard = (props: {
                     ) : (
                         <div className='Location'>Online</div>
                     )}
-                    <img
-                        src={CalendarExport}
-                        alt='Export to calendar'
-                        className='CalendarExportIcon'
+
+                    {numAhead > 0 ? (<div className={'Indicator ' + props.status}>
```
rgu0114, line 97: similar code style thing here: use `condition && component` instead of `condition ? component : null`

rgu0114 (changes requested):
Great work here Nidhi! Agree with Steven and Lily's comments. The functionality of the redesign works as intended, and the ui matches the designs closely (very clean btw). I also liked how comprehensively you tested this + listed out the steps. Aside from a couple nit pick code style changes, could you attach screenshots of the before vs. after and the figma designs for future reference? Other than that, should be ready to merge soon!

rgu0114 (approved):
👍

PR #863, opened 2024-05-04
Professor Student View + Student Leave Queue Bug Fix

stevenyuser:
### Summary 

Adds professor student view, a feature where professors can view the queue as a student (and do all the things a student could do e.g. add themselves to the queue).
- Once a professor adds themselves to a queue as a student, they are treated like a student inside of professor student view
- In the normal SplitView (Queue tab), they still have all the permissions of a normal professor
  - includes the ability for a professor (or TA) to assign themselves to any queue question (including their own question)
- Introduces new prop in SessionView and its children to showProfessorStudentView (overrides a potential Professor/TA's roles to display them as a student)
- Does not affect functionality of normal SplitView (Queue tab)

image
image
image

Also fixes a bug which prevented students from removing themselves from the queue
- Bug was caused by the Feedback feature, which didn't pass and set the removeQuestionId in SplitView.tsx properly
- Fix by introducing new separate functions/props for setRemoveQuestionId (removes a student's question from the queue) and removeQuestionDisplayFeedback (used to display feedback prompt when a student's question is done)

### Test Plan 

ProfessorStudentView
- Tested as a professor
  - Can add self to queue, can view and do everything like a student would in SplitView
  - Can remove the professor's question from the queue
  - In SplitView, professor can assign and answer questions just as before
- Tested as a student to ensure SplitView still works as normal

Leave Queue Bug Fix
- Tested leaving queue as a normal student
- Tested leaving queue in ProfessorStudentView as a professor

### Notes 

- There might be a possible bug in ProfessorStudentView with qmijinglefinal.mp3, because I see in the console that it cannot be found (HTTP 500 error code), so I think ProfessorStudentView can't play that audio. I think this has something to do with directories. Before, SessionView was inside SplitView only, but now it is in ProfessorStudentView, which has a different directory structure than qmijinglefinal.mp3. However, this bug doesn't really affect the core functionality of ProfessorStudentView.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


NIDHI2023 (commented):
This works really well! I see that you considered the view for mobile users too, which I think is good forethought. Removing and adding my own question in the student view also seems to be working.

One thing I'm a little unsure of, is there a clearer way for the user to know when they are leaving Professor/Student view? Right now I think the pages look very similar visually, so it could benefit having a simple outline (like Sia's designs) around the student view in the ProfessorStudentView page

Also minor note:
On line 146 in ProfessorStudentView.tsx and line 152 in SplitView.tsx I don't think a console.log call is needed anymore. I think that it's good to avoid having prints to the console for the "finished" version once you're done debugging or developing.

rgu0114:
> This works really well! I see that you considered the view for mobile users too, which I think is good forethought. Removing and adding my own question in the student view also seems to be working.
> 
> One thing I'm a little unsure of, is there a clearer way for the user to know when they are leaving Professor/Student view? Right now I think the pages look very similar visually, so it could benefit having a simple outline (like Sia's designs) around the student view in the ProfessorStudentView page
> 
> Also minor note: On line 146 in ProfessorStudentView.tsx and line 152 in SplitView.tsx I don't think a console.log call is needed anymore. I think that it's good to avoid having prints to the console for the "finished" version once you're done debugging or developing.

+1 on Nidhi's comment related to the visual distinction between Professor/Student view

rgu0114 (approved):
Awesome work with this Steven! I remember us discussing how to approach this last semester, and it looks like you've been pretty successful with the implementation + you handled the role switching in a smart way. I tested all the functionality and everything looks good to me, so I'll approve this for now, pending some code style nit picks.

src/components/App.tsx
```diff
@@ -8,44 +8,42 @@
...
-import * as React from 'react';
-import { RouteComponentProps } from 'react-router';
-import { BrowserRouter as Router, Route, Redirect, Switch } from 'react-router-dom';
-import * as ReactGA from 'react-ga';
-import { Loader } from 'semantic-ui-react';
+import * as React from "react";
+import { RouteComponentProps } from "react-router";
+import { BrowserRouter as Router, Route, Redirect, Switch } from "react-router-dom";
+import * as ReactGA from "react-ga";
+import { Loader } from "semantic-ui-react";
```
rgu0114, line 15: hmm, seems like some pretty big diffs resulting from your linter. Ideally would be nice if you could remove those changes from the PR but it's fine if not.

rgu0114 (commented):
Also thanks for fixing the remove student from queue bug from the Feedback feature, good catch haha

src/components/includes/SessionView.tsx
```diff
@@ -51,43 +55,58 @@ type RouteParams = {
...
+    user,
+    setShowModal,
+    setRemoveQuestionId,
+    removeQuestionDisplayFeedback,
+    timeWarning,
+    sessionBanners,
+    showProfessorStudentView,
+}: Props) => {
+    // make user appear as not a ta/prof if showProfessorStudentView is true
+    const isTa = showProfessorStudentView ? false : user.roles[course.courseId] !== undefined;
```
rgu0114, line 74: nit: `!showProfessorStudentView && user.roles[course.courseId] !== undefined;` might be cleaner

src/components/pages/ProfessorStudentView.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    />
+                )}
+                <CalendarExportModal
+                    showCalendarModal={showCalendarModal}
+                    setShowCalendarModal={setShowCalendarModal}
+                    isDayExport={isDayExport}
+                    currentExportSessions={currentExportSessions}
+                    course={course}
+                />
+                {(width > MOBILE_BREAKPOINT || activeView !== "calendar") &&
```
rgu0114, line 177: hmm, I think there's quite a lot of code redundancy between this new file and SplitView, but that can be refactored in a future PR

PR #868, opened 2024-09-18
Fix Undefined TA Link Update

rgu0114:
### Summary 

Provide a non-undefined string fallback for TALink. Same fix as [this PR](https://github.com/cornell-dti/office-hours/pull/818) but for the useEffect TALink update, in addition to just the state initialization.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


NIDHI2023 (approved):
Good catch, this can prevent future issues with undefined values!

PR #869, opened 2024-09-22
QMI Wrapped Intro Animation

NIDHI2023:
### Summary 

Created a temporary button on top navbar to show a modal with the QMI Wrapped intro animation.

- imported 3 new media files (arrow.svg, plus.svg, plus2.svg)
- created a new folder in /styles for qmi wrapped
- added comments explaining how animations work

https://github.com/user-attachments/assets/0697b245-103f-4856-9389-a3578dd4e913

### Test Plan 

- check if modal popup and animation looks good/consistent for different device screens

### Notes 

- need to make folder and style changes consistent with Richard's setup for wrapped frontend so it's easier to build from this modal
- could probably get rid of 2nd plus and just go into a consistent blue background, from there apply a gradient according to the Figma designs

### Breaking Changes 

None

- My changes requires a change to the documentation.
  - I added an extra style folder, so that would need to be updated. Depending on adjustment to [Richard's PR setup](https://github.com/cornell-dti/office-hours/pull/854), would not be needed

### Checklist

- [ ] My PR adds a @ts-ignore


rgu0114 (commented):
Excellent work here with the intro animation Nidhi! It's exciting to see Sia's Figma designs slowly come to life, and I'm looking forward to getting this to prod. Great job figuring this out by yourself and for leaving helpful comments explaining how it works haha

Visually the only thing I'd make note of is the speed of the animation – we can sync up on that during the next couple weeks with the designers.

Since the [Wrapped Frontend setup](https://github.com/cornell-dti/office-hours/pull/854) PR has been merged, we should be good to add this animation in to the beginning of the slideshow modal. And in that PR, we created a media/wrapped folder so remember to add the svgs here to that.

src/components/includes/WrappedAnimationModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+                                >
+                                    <stop stopColor="#FF9399" />
+                                    <stop offset="1" stopColor="#FF5A60" />
+                                </linearGradient>
+                                {/* this is the actual circle part. 
+                                cx and cy are centers so shld be half of height/width. r is radius */}
+                                <circle cx='150' cy='150' r='115'> </circle>
+                            </svg>
+                            <svg className="blue-circle" width="400" height="400">
+                                <defs>
```
EricWeng23, line 57: I was running this locally and the modal works amazingly, but I'm just curious why in the blue circle svg there is a <defs> tag versus not having one for the red svg. I was wondering if this was a particular design choice on intention. iirc the <defs> element is used to store graphical objects and used when we want to render the particular element. The modal looks very good visually :)

EricWeng23 (commented):
The animation modal looks very good viusally and reflects the figma very well. It's just a small minor nitpick of a particularly `<def>` tag in the `WrappedAnimationModal` component in which I already commented in that code block. But this was amazing implementation to make the modal so dynamic :)

src/components/includes/TopBar.tsx
```diff
@@ -34,6 +37,14 @@ const TopBar = (props: Props) => {
     const [showMenu, setShowMenu] = useState(false);
     const [showTextModal, setShowTextModal] = useState<boolean>(false);
     const [image, setImage] = useState(props.user ? props.user.photoUrl : "/placeholder.png");
+    const [showWrappedModal, setShowWrappedModal] = useState<boolean>(false);
+
+
+    const whoops = () => {
```
anchen9, line 43: Hi Nidhi! The animation looks really cool and I agree with Richard on the speed of the animation! I was just curious about the name of this function and if it could be more descriptive? (I love it though :D) Thank you so much for also including comments throughout!

NIDHI2023: Oh I was just trying to see if it worked and totally forgot to change it haha, thanks for the catch!

src/components/includes/WrappedAnimationModal.tsx
NIDHI2023: Also good catch! I believe both linear gradients are used in the same way so it would be more consistent to have
`<defs>` tags for both of them.

PR #875, opened 2024-10-27
Loading Screen to Welcome Screen

NIDHI2023:
### Summary 

Building on [this Wrapped Frontend PR](https://github.com/cornell-dti/office-hours/pull/854) and this [animation PR](https://github.com/cornell-dti/office-hours/pull/869), creates a loading animation screen that transitions into the welcome screen.

https://github.com/user-attachments/assets/bf624ea5-5de3-4869-b902-a3ef7349edf4

- Made animation faster
- Fixed some centering issues on different devices

### Test Plan 

- Loading animation should look centered on different device widths
- No lags or delays, and dots and arrows should only show up after the welcome component is rendered

### Notes

- I made the loading animation have a 0.8s delay because shortening this caused some odd CSS flickering, and this was the lowest time I could go without the flickering

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/Wrapped.tsx
+++ b/src/components/includes/Wrapped.tsx
@@ -102,6 +102,13 @@ const Wrapped= (props: Props): JSX.Element => {
         </div>
     );
 
+    const handleAnimationEnd = () => {
+        const timer = setTimeout(() => {
+            setLoading(false);
+        }, 1000); 
+        return () => clearTimeout(timer);
+    }
+
     const WrappedAnimationModal = () =>
         (                 
             <div className='qmi-container'>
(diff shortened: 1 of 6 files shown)


rgu0114 (changes requested):
Great work integrating the animation into the transition into the Wrapped modal. Since we have this, I think we can close the original animation PR. I did notice from testing on my side though that there was a bit longer of a delay (a few seconds) where it seemed a little stuck on the DTI logo. Are you seeing this as well? Could just be slow network speed on my side.

Screen Shot 2024-10-27 at 4 12 32 PM

src/components/includes/Wrapped.tsx
```diff
@@ -86,39 +93,114 @@ const Wrapped = (props: Props) => {
...
+                <img src={arrow} className="arrow-circle" alt="dti arrow" />
+                {/* made two pluses since color gradient changes and second one needs
+               to expand outside of the container. */}
+                <img src={smallPlus} className="first-plus" alt="first plus" />
+                <img
+                    src={bigPlus}
+                    className="sec-plus"
+                    alt="second plus"
+                    onAnimationEnd={() => {
+                        const timer = setTimeout(() => {
```
EricWeng23, line 160: Overall, the animation looks good. I'm just curious that since this calls an inline-function inside of `onAnimationEnd`, wouldn't this create a timer every re-render. Would it be better to make declare the function outside? This might be a bit better performance-wise if the component frequently re-renders or updates.

rgu0114 (approved):
good to merge once changes are addressed

PR #878, opened 2024-10-28
Wrapped Countdown Component

EricWeng23:
### Summary 

Created a countdown for the QMI Wrapped release start 10 days before launch. The page will render a ribbon ball at the button right that will open the countdown when clicked. The countdown component consist of 3 counter: Days, Hours, Minutes. Note that the ribbon ball and countdown component will NOT be visible until the start date for which QMI Wrap begins its countdown. Unfortunately, currently the confetti for when counter reaches 0 is not yet implemented, thus only a cone svg is visible as of right now. Once counter reaches 0, user will be prompted to click a button to view their QMI Wrapped slider which is the `Wrapped` component.

To keep track of the varying start date and launch date across different semester, update the dates inside of the
`constants.ts` file

### Test Plan 

Tested via the local server to check if component renders and functions as expected. Pictures below:

**When today is NOT the start date, so no component will be rendered**
Screenshot 2024-10-27 at 7 18 01 PM

**When today is the start date, then the counter component will be displayed with DAYS, HOURS, MINUTES
Note this is just a test example for visual purposes, so the 18 days depicted is not what will be shown on start date**
Screenshot 2024-10-27 at 7 19 57 PM

**When the counter is 0, then user will be prompted to click on the view button to open the Wrapped component
Note this is just an example of what happens when counter is 0**

https://github.com/user-attachments/assets/adb5d3a6-cb65-487f-8c2c-a602097f99e5

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/components/pages/SplitView.tsx
```diff
@@ -274,16 +285,21 @@ const SplitView = ({
                     <Loader active={true} content="Loading" />
                 ))}
             <ProductUpdates />
+            {!isStartDate() ? null : countDownClicked ? (
```
rgu0114, line 288: convention to simplify this to `isStartDate() && ...` instead of conditionally returning null

src/components/pages/SplitView.tsx
```diff
@@ -198,6 +202,13 @@ const SplitView = ({
         }
     }, [addBanner, user]);
 
+    // Check that today is the start date, then render the countdown if true
+    const isStartDate = () => {
```
rgu0114, line 206: shouldn't we instead be displaying `<WrappedCountdown />` anytime the current date is after the start date, instead of just on the start date?

src/components/includes/WrappedCountdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+                            <p className="top">Queue Me In</p>
+                            <p className="bottom">WRAPPED</p>
+                        </div>
+                    </>
+                ) : (
+                    <>
+                        <button type="button" className="viewWrap" onClick={handleButtonClick}>
+                            View
+                        </button>
+                        <img className="textContainer cone" src={cone} alt="icon" />
```
rgu0114, line 83: oh I think the cone is actually for the confetti to shoot out of, and actual clickable view to display the wrapped model looks like this:
Screen Shot 2024-10-30 at 1 24 22 AM
let's confirm this during work session tomorrow. Since the confetti is lower priority, let's actually only include the clickable view for now and maybe add the confetti in a follow-up PR, since it'll only show up for a few seconds.

src/components/includes/WrappedCountdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+// Declare interface to takes in the setter setDisplayWrapped as a prop
+interface WrappedCountdownProps {
+    setDisplayWrapped: React.Dispatch<React.SetStateAction<boolean>>; // Define prop type for setter
+}
+const WrappedCountdown: React.FC<WrappedCountdownProps> = ({ setDisplayWrapped }) => {
+    const handleButtonClick = () => {
+        // Call the setter function to change the state in the parent
+        setDisplayWrapped(true); // Set to true to show the modal
+    };
+    const launch = new Date(WRAPPED_LAUNCH_DATE);
```
rgu0114, line 24: Date object creation is redundant each render. Can we move these outside this component, maybe to SplitView?

src/components/includes/WrappedCountdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const start = new Date(WRAPPED_START_DATE);
+    const date: WrappedDate = {
+        launchDate: launch,
+        startDate: start,
+    };
+    // Define a zeroCounter object for comparison when there's no time remaining
+    const zeroCounter = { days: 0, hours: 0, minutes: 0 };
+    // Helper function to calculate the remaining time in days, hours, and minutes from start date to launch date
+    const calculateTimeRemaining = (dateProps: WrappedDate): RemainingTime => {
```
rgu0114, line 33: nit: utility functions like these should be declared outside the component, near the top of the file

rgu0114: also it might be better to define some of these values for time calculations (like 1000 * 60 * 60 * 24) as constants to improve readability

src/components/includes/WrappedCountdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+        // Calculate days, hours, and minutes remaining
+        const days = Math.floor(gap / (1000 * 60 * 60 * 24));
+        const hours = Math.floor((gap / (1000 * 60 * 60)) % 24);
+        const minutes = Math.floor((gap / (1000 * 60)) % 60);
+        return { days, hours, minutes };
+    };
+    // Initialize countdown state using the calculateTimeRemaining function
+    const [timeRemaining, setTimeRemaining] = useState<RemainingTime>(calculateTimeRemaining(date));
+    useEffect(() => {
+        const interval = setInterval(() => {
```
rgu0114, line 45: Nice job setting this up to update the time remaining every second!

rgu0114 (commented):
Looks great so far Eric! Nice job figuring out the countdown logic independently and styling the component from scratch. It looks pretty good from the pictures and the screen recording you attached. It's not displaying for me when I run it locally for some reason, but we can sync on this during work session tomorrow. 

The main thing to note for next steps is the transition to the clickable view (refer to [Figma](https://www.figma.com/design/gKqRwgffXSKNB5PYQlXHS0/Sia-FA'24-Working-File?node-id=1-6&node-type=canvas&t=vAuiNSeLmwe8BfFf-0) again), which will be a separate step after animating the confetti for some duration. Other than that though, it's looking promising and I'm excited to see this launch!

rgu0114 (commented):
Here's what I'm thinking: We launch the countdown component with the countdown itself and the click to view modal when countdown reaches zero first (probably after Nov. 5th). Then, we work on the confetti in a separate PR during that 10 day countdown so it's not blocking us on anything.

rgu0114 (approved):
Works great on my side! Confetti's exciting to see. Approving, but also make sure to update the PR description before merging.

PR #880, opened 2024-11-01
Minor Tag Info UI Fixes

rgu0114:
### Summary 

- fixed add new tag input field at the top, under "Tags" and above the persisted list of new tags
- added extra margin so that text field and selected child tags don't get squeezed together

Screenshot of the resulting fixes:
Screen Shot 2024-11-01 at 12 47 16 PM

Previously:
Screen Shot 2024-11-01 at 12 49 37 PM


NIDHI2023 (approved):
Works great for me!

PR #887, opened 2024-11-07
edit classes frontend redesign

swang235:
### Summary 
Update of edit classes page:

current design:
image
image

redesign:
image
image

### Test Plan

- testing different cases (ex class pairings, roles, search bar)
- tested responsiveness for ^ as well

### Breaking Changes 

None

- [x ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore


NIDHI2023 (commented):
This is a really nice redesign! It looks so much more cleaner. I have a few suggestions you could add:
- When searching for a class that doesn't exist, maybe add some padding between the label thing and the end of the white part so it doesn't look too crammed together

image

- I think making the Go to Course buttons more visible by changing the color or adding a border or something else would be good. And also maybe adding more padding again to make the class cards look more vertically symmetric 

image

rgu0114 (changes requested):
Screen Shot 2024-12-05 at 2 33 08 PM

It seems like only non-prof courses have the hover shadow effect. Is this intended? The prof courses are clickable but don't have any effect.

rgu0114 (changes requested):
I think the font of the search bar placeholder text can also match more closely
Screen Shot 2024-12-05 at 4 56 03 PM
Screen Shot 2024-12-05 at 4 56 19 PM

src/components/includes/CourseCard.tsx
```diff
@@ -11,53 +12,103 @@ type Props = {
...
     };
 
-    let roleString = '';
-    if (role === 'ta') {
-        roleString = 'TA';
-    } else if (role === 'professor') {
-        roleString = 'PROF';
+    let roleString = "";
+    let roleColor = "";
+    let selectedBackgroundColor = "#F5F5F5";
```
rgu0114, line 42: oh can we put these in the associated .scss file?

src/components/includes/CourseCard.tsx
```diff
@@ -11,53 +12,103 @@ type Props = {
...
+                                        <Icon className="icon" color="grey" name="circle outline" />
+                                    )}
+                                </div>
+                            </Grid>
+                        ) : (
+                            <Grid container item xs={6} justify-self="end" />
+                        )}
+                    </div>
+                ) : (
+                    <></>
```
rgu0114, line 99: nit: let's avoid empty components `<></>` and use `&&` where we can, assuming that has the same behavior

src/components/includes/CourseCard.tsx
```diff
@@ -1,6 +1,7 @@
-import * as React from 'react';
-import { useHistory } from 'react-router';
-import { Icon } from 'semantic-ui-react';
+import * as React from "react";
+import { useHistory } from "react-router";
+import { Icon } from "semantic-ui-react";
+import { Grid } from "@material-ui/core";
```
rgu0114, line 5: General: non prof/ta former classes still seem to have the hover effect (benji 2110) when idt they should
Screen Shot 2024-12-06 at 1 44 37 AM

rgu0114 (changes requested):
just a couple more minor things, everything else looks good!

src/components/includes/CourseCard.tsx
laylaliuuu: There are a few lines that I'm unsure if it's intentional or could be repetitive, what does selected-background-color do, and why does only student role have this background in the if else: 

 .scss:
   selectedBackgroundColor-role-student: "rgba(214, 234, 254, 0.4)"; 
    selectedBorderColor-role-student: "#77BBFA";

    &.selected {
        background-color: rgba(214, 234, 254, 0.4);
        border: 2px solid #77BBFA;
    }
    
    .tsx:
if (role === "ta") {
        roleString = "TA";
        roleColor = "#var(role-color-ta)";
    } else if (role === "professor") {
        roleString = "PROF";
        roleColor = "#var(role-color-professor)";
    } else {
        selectedBackgroundColor = "selectedBackgroundColor-role-student";
        selectedBorderColor = "selectedBorderColor-role-student";
    }

PR #895, opened 2024-11-14
Aggregate Wrapped Data - updated version

NIDHI2023:
### Summary 

Made lots of edits to [base code](https://github.com/cornell-dti/office-hours/pull/874). 
- Added new fields for favorite month, favorite day, and number of students helped for TAs
- Changed to favorite courseId instead of sessionId and got rid of favorite title
- Refactored code to make more readable
- Added some optimizations when getting sessionDocs and userDocs by calling Promise.all first instead of calling await every time in a loop
- Added progress debugging statements for when we run it on bigger data, so we know where the script is at

### Test Plan 

- New fields have correct values in test data
- user in wrapped who’s not found in userdocs does not get wrapped:true
- user who isn’t in wrapped does not get wrapped:true

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,7 +2,7 @@ name: CD
 on:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   deploy:
--- a/README.md
+++ b/README.md
@@ -117,5 +117,5 @@ _Screenshots showing major parts of app_
 
 
 We are a team within **Cornell Design & Tech Initiative**. For more information, see our [website](https://cornelldti.org/).
-<img src="https://raw.githubusercontent.com/cornell-dti/design/master/Branding/Wordmark/Dark%20Text/Transparent/Wordmark-Dark%20Text-Transparent%403x.png">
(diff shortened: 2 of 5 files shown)


rgu0114 (approved):
Everything seems logically correct to me! Great work with this over the past couple of months and handling complex cases with fav TA and stuff. I'm still a bit nervous about the scale of reads/writes we're running, so one option would be to subdivide the date ranges even more, but we can figure that out on Monday.

.github/workflows/cd.yml
```diff
@@ -2,7 +2,7 @@ name: CD
 on:
   push:
     branches:
-      - master
+      - main
```
rgu0114, line 5: oh also prob don't include this change and the one in the readme

rgu0114 (changes requested):

anchen9:
Hi Nidhi! Thanks for adding the fields for favorite day, favorite month, and number of students helped! The logic for favMonth makes sense and I like that you returned them as numbers since it was very intuitive to incorporate! Your comments throughout are also very helpful when looking over your code!   I did notice a few commented out console statements that maybe can be removed?

PR #879, opened 2024-10-28
Wrapped Frontend - Add TA slides  & Update slide logic

anchen9:
### Summary 

Added three total slides to the Wrapped Frontend by building on this [PR](https://github.com/cornell-dti/office-hours/pull/854) and Figma dev handoff. I added slides is to display a student's favorite TA, the total time a TA helped students, and a slide for how many students a TA has helped.  Also updated slide logic so that if an user is only a student, they do not see TA slides, and if an user is both a student & TA they see all slides, and if an user is only a TA they only see TA specific slides.

- Added background gradient for all wrapped slides
- Flipped image on the visits to office hours slides
- Edited image size of personality slides
- Edited text appearance on conclusion slide
- Added favTaId to wrappedData 
Screenshot 2024-11-17 at 12 18 15 PM
Screenshot 2024-11-17 at 12 18 23 PM
Screenshot 2024-11-17 at 12 18 30 PM
Screenshot 2024-11-17 at 12 18 37 PM
Screenshot 2024-11-17 at 12 18 43 PM
Screenshot 2024-11-17 at 12 18 49 PM
Screenshot 2024-11-17 at 12 19 24 PM
[
Screenshot 2024-11-17 at 12 42 04 PM
](url)

### Test Plan 

To test the slide logic, I edited my test data in QMI Test to test:

- favTaId = "" or undefined  ==> should only see TA specific slides
- timeHelpingStudents == 0 or undefined ==> should only see student specific slides
- if both are defined and non-empty/non-zero then all slides should be displayed.

### Notes 

- need to complete ribbons/banners as well as animations
- need to add firebase call to number of students a TA has helped.

### Breaking Changes 

None

- [ ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore

PR #897, opened 2024-11-17
Wrapped Frontend - Adding People to Numbers

anchen9:
### Summary 

Built on this [PR](https://github.com/cornell-dti/office-hours/pull/879) by adding the people on the numbers on the slide with time spent at office hours and the TA slide of number of students helped. Each number is its own svg and different numbers of digits have different rules on where the people are placed.

- fixed == to === everywhere

Screenshot 2024-11-17 at 12 20 16 PM
Screenshot 2024-11-17 at 12 43 46 PM

### Test Plan 

I tested numbers of each length, ex. one digit numbers, five digit numbers, etc. I also tested different combinations of digits so that I could see how each digit appears in different spots. 

### Notes 

- Some numbers look slightly non-uniform, likely just due to the whitespace in the svg, but this should be a quick fix by redownloading!

### Breaking Changes 

None

- [x ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore


rgu0114 (approved):
lgtm! let's demooo

PR #899, opened 2024-11-18
Wrapped ribbon notification

EricWeng23:
### Summary 

Added ribbon to notification bell after counter reaches 0 or in other words past launch date. This required creating passing
a prop `countdownZero` to the `TopBar` component which passes it down to its child component `TopBarNotifications`. Depending on the value of `countdownZero` it will render the appropriate bell svg as well as a notfication message indicating the launch of the user wrapped statistics.

### Test Plan 

Tested via local server to check behavior of the notification bell

https://github.com/user-attachments/assets/86e5e2c6-ebe1-4f13-abe7-40d2a428b9a4

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rgu0114 (approved):
yay looks great! Looks like ur getting this linting error tho 
Error:    19:5   error    propType "countdownZero" is not required, but has no corresponding defaultProps declaration     react/require-default-props

I think u can fix that with smtg like the following:
`TopBarNotifications.defaultProps = {
    countdownZero: false,
};
`
below ur component

PR #901, opened 2024-11-21
Testing + small css fixes

NIDHI2023:
### Summary 

- fixed bug with how wrappedData was stored from collection, it should now show correct numbers
- changed "semester" words to "year" in ui
- added white hover to dots to make it easier to see you can click
- fixed hover on arrows to resemble the close icon

### Notes 

- If we're considering mobile users, anything below 800 px you can't easily exit Wrapped or click to other slides

800px:
image

700px:
image

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/Wrapped.tsx
+++ b/src/components/includes/Wrapped.tsx
@@ -201,7 +201,7 @@ const Wrapped= (props: Props): JSX.Element => {
     useEffect(() => {
         // add usestate for totalstages calculate it in useEffect
 
-        const wrappedRef = firebase.firestore().collection('wrapped-fa24');
+        const wrappedRef = firebase.firestore().collection('wrapped');
         const fetchData = async () => {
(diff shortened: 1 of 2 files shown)


rgu0114 (approved):
Good note abt the mobile users, we'll need to patch that up in a future pr

rgu0114 (changes requested):
oh remember to pull from `wrapped` and not `wrapped-fa24`

swang235 (approved):
i agree w richard about wrapped-fa24 but otherwise lgtm! good catch on mobile view and nice work nidhi!

NIDHI2023:
i haven't tested this on prod database (since it wasn't working for me) so I think this should be tested on prod before merging

rgu0114 (approved):
lgtm!

PR #893, opened 2024-11-12
Deploy Wrapped Countdown

rgu0114:
### Summary 

Deployment of [this PR](https://github.com/cornell-dti/office-hours/pull/878)

### Test Plan 

Typical QMI functionality working

PR #902, opened 2024-11-21
Countdown Component Fixed Minor Bugs

EricWeng23:
### Summary 

This is a follow-up PR since the last deployment of the countdown PR <a>https://github.com/cornell-dti/office-hours/pull/893</a> that addressed minor bugs listed below:
- Add an extra check `user.wrapped`, so countdown only gets rendered if wrapped data exist
- Fixed the grammatical singular and plural form of DAY(S), HOUR(S), MINUTE(S)
- Allow wrapped notification to be clickable to open wrapped modal
- Previously having `timeRemaining` in the dependency array of `useEffect` to update the countdown results in an infinite render, thus added a timer that will update the countdown every 1 second interval

### Test Plan 

Tested the changes on local server

Screenshot 2024-11-21 at 5 31 58 PM
e changes on local server

https://github.com/user-attachments/assets/8a5fe137-2b40-44ea-b9e4-fd60d5fae901

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/TopBar.tsx
+++ b/src/components/includes/TopBar.tsx
@@ -29,6 +29,7 @@ type Props = {
     admin?: boolean;
     snackbars: Announcement[];
     countdownZero?: boolean;
+    setDisplayWrapped?: React.Dispatch<React.SetStateAction<boolean>>;
 };
 
 const TopBar = (props: Props) => {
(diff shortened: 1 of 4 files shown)


rgu0114 (commented):
oops I accidentally also did the hasWrapped check for countdown, i'll remove it on my pr but I did the notif bell as well. nice catch with the infinite render – noticed that yesterday on dev but not prod but forgot to post abt that.

rgu0114 (approved):
Changes look good!

PR #913, opened 2024-12-04
Wrapped emailing

NIDHI2023:
### Summary 
Credit to [CoursePlan's email system](https://github.com/cornell-dti/course-plan/tree/main/scripts/email) for instructions to set up the system.

This PR creates a script that uses the Resend API to send automated emails to users that can see Wrapped. At most, it can send 100 emails each as a batch with 49 recipients in a day.

If it's the first time running the script, you should run: `node <script path> 0`
This will start the script from the first email (0th index). If the 100 email limit has been hit, there will be a console statement that tells you the next input number to run the next day.

- ran npm install dotenv
- ran yarn add resend

### Test Plan 

- Throws error and stops control before calling the API if the emails exceed limit
- BCC recipients cannot see each other, but they can see the main "TO" email - this should be a dummy email or the general DTI email
- Images across different screens:
Laptop:
image

Mobile:
image
image

### Notes 

- In the prod database would need to also update indexes for user like below:
image
Firebase auto-generates this when you run the script and get an initial error message with a link:
image
Should look like this at the end and script should be working
image

- Created online references to images with ImgBB

- Fonts like Roboto don't seem to be rendering properly across all email providers - might just not be easily feasible

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/scripts/wrapped-email.ts
```diff
@@ -0,0 +1,44 @@
...
+import admin from "firebase-admin";
+
+admin.initializeApp({
+    credential: admin.credential.applicationDefault(),
+    databaseURL: 'https://qmi-test.firebaseio.com'
+
+});
+
+// i think api key goes in here
+// const resend = new Resend('re_123456789');
```
rgu0114, line 11: o remember to stash this in a .env if this is your resend api key

rgu0114 (changes requested):

src/scripts/wrapped-email.ts
NIDHI2023: it's not, just a default test!

rgu0114 (commented):
Nice job testing this thoroughly and documenting your process well

rgu0114 (commented):
Great work with this and figuring out the Resend API! No glaring issues I can see with this – only thing to watch out for when we run it is if we hit a limit, what's the best way to continue where we left off (if manually doing it is the best option, that's ok too).

PR #914, opened 2024-12-06
Add parameter to NumberPeople

anchen9:
### Summary 

NumberPeople originally used only totalMinutes as digits, but this resulted in inaccurate numbers being reflected in TaStudentsHelped. A number parameter was added to NumberPeople so that the correct number could be inputted. 

Screenshot 2024-12-05 at 9 11 59 PM
Screenshot 2024-12-05 at 9 12 23 PM

### Test Plan 

Changed Firebase values for numStudentsHelped and totalMinutes to be different. 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rgu0114 (approved):
yep looks good!

PR #915, opened 2024-12-06
Add parameters to NumPpl - Release

anchen9:
### Summary 

Add parameters to NumberPeople so that different numbers can be inputted rather than defaulting to totalMinutes.

Screenshot 2024-12-05 at 9 11 59 PM
Screenshot 2024-12-05 at 9 12 23 PM

### Test Plan 

Made totalMinutes and numStudentsHelped different.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


rgu0114 (approved):
lgtm! this should fix the issues with the incorrect values appearing in how many students helped by TAs

PR #921, opened 2025-01-24
Calendar export redesign css fixes

NIDHI2023:
### Summary 

Small css fixes for #862 , especially mobile view
- changed dot indicator to actually be red dot, not flash green when TA active
- removed long fadeIn time for mobile modal view
- adjust "export to gcal/ical" text font size for mobile view
- added hover effect to selecting calendar dates to match design prototypes

https://github.com/user-attachments/assets/56443c82-a067-446f-8e65-d456a1d77e5a

### Test Plan 

- red dot only appears for TAs, not students, instead of previously styled indicator when there is a question in the queue
- could test with more than one question if it still works
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,15 @@
 # dependencies
 node_modules
 
+# yarn files
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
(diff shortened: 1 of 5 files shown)


EricWeng23 (approved):
Looks good to me!

PR #923, opened 2025-02-08
Eric / Implemented Script For Automating Email Outreach Information (In Progress)

EricWeng23:
### Summary 

*UPDATE 2*:
Added the additional column 'Has course/professor use QMI before' to the output CSV file 
- Populated the column cells using data from past semester tracker csv (SP24)
   - Past semester csv located in `scripts/tracker` folder called `past_semester.csv`
- Blacklisted courses that are not interested in using QMI

As documented the code for the status of columns are as follows:
- 2 indicates that both the course and professor have used QMI before
- 1 indicates that professor used QMI before
- 0 indicates that course used QMI before
- -1 indicates that neither the course nor the professor have used QMI before
    
Implementation Logic:
1) If one of the professors teaching this semester has taught the course in the past, check whether the course decided to use QMI that past semester
    - If the course decided to use QMI, return 2 (since both professor and course used QMI)
    - If the course decided not to use QMI, then return the original status
2) If a new professor teaching the course this semester, check if the course previously used QMI
    - If the course has decided to use QMI in the past semester, return 0 (course has used QMI, maybe not prof)
    - If the course decided not to use QMI, but historically the course or the past professors used QMI (0 or 2) return 0

Return -1 if the course is not tracked in spreadsheet.
Screenshot 2025-02-22 at 8 45 59 PM

*UPDATE*: 
Additional Filtering Logic implemented:
1) Filtered out Academic Support
2) Filtered out Independent Research and Studies
3) Filtered out Graduate Level courses
4) Filtered out Cornell Tech Courses
5) Filtered out courses with < 3 credits
I still kept the bar for class level for <5000 because some 4k classes still has high demand for OH. Filtered out the Independent Research and PRJ instead.

From Nidhi's Comment regarding the Professor teaching multiple courses, it would happen under 2 cases:
1) Professor ACTUALLY teaches 2 courses
2) A course is crosslisted with other subject fields
A check is implemented to combine these courses into one like below:
Screenshot 2025-02-18 at 12 01 07 AM

This PR focuses on implementing a script to automate and compile all the course informations (Course Number, Professors, Emails, etc) into a csv file. The data are fetched from the class roster API to retrieve the JSON files for all the relevant courses that QMI is tracking.

API URL: https://classes.cornell.edu/api/2.0/config/rosters.json
*Note*: The semester rosters are sorted in chronological order with the last element being the most recent SP25
Screenshot 2025-02-08 at 3 30 27 PM

Supply the appropriate URL parameters to get the corresponding data (e.g specify the subject)
Screenshot 2025-02-08 at 3 33 49 PM
Screenshot 2025-02-08 at 3 36 02 PM

**NOTE**: Class roster API does not supply the total class size of the course, so we are unable to filter with this property. Researched Course-Grab github repo, but its seems like they are only fetching the status code (open, closed, waitlist) of the courses instead of class sizes. Thus, course grab does not apply to our use case. An alternative is to webscrape from Student Center, but this is extremely difficult to implement as I had experimented already. 

This only compiles the csv with the subjects courses we discussed, did not implement much of the filtering logic (only filtered out some graduate courses)

### Test Plan 

Tested locally via the terminal and verified that contents are written to the output csv file
Screenshot 2025-02-08 at 3 43 03 PM

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/scripts/course-scrape.py
```diff
@@ -0,0 +1,76 @@
...
+]
+
+classesJSON = []
+for s in subjects:
+  if s['value'] in relevent_courses:
+    classURL = "https://classes.cornell.edu/api/2.0/search/classes.json?roster="+roster+"&subject=" + s['value']
+    classesJSON.extend(grab_json(classURL)['data']['classes'])
+
+def getProfessorName(instructors):
+  accum_names = "" # To accumulate all the professor names
```
anchen9, line 43: Hi Eric! Great work so far on implementing the script for email automation! Your organization throughout and consistent informative comments which make understanding your code a much smoother process!

src/classes.csv
```diff
@@ -0,0 +1,270 @@
...
+CHEM 3030,H. Floyd Davis,hfd1@cornell.edu
+CHEM 3570,Cynthia Kinsland,clk10@cornell.edu
+CHEM 3580,"Geoffrey Coates, Brett P. Fors","gc39@cornell.edu, bpf46@cornell.edu"
+CHEM 3590,Bruce Ganem,bg18@cornell.edu
+CHEM 3880,Barbara A. Baird,bab13@cornell.edu
+CHEM 3900,Greg Ezra,gse1@cornell.edu
+CHEM 4100,Peter Thomas Wolczanski,ptw2@cornell.edu
+CHEM 4210,Kyle M. Lancaster,kml236@cornell.edu
+CHEM 4330,Brian Crane,bc69@cornell.edu
+CHEM 4430,"Peng Chen, Prof.",pc252@cornell.edu
```
NIDHI2023, line 39: This is an unusual case here, maybe you could check for that?

src/classes.csv
```diff
@@ -0,0 +1,270 @@
...
+INFO 4400,Gilly Leshed,gl87@cornell.edu
+INFO 4410,Malte F Jung,mfj28@cornell.edu
+INFO 4450,Susan R. Fussell,srf72@cornell.edu
+INFO 4900,Yian Yin,yy994@cornell.edu
+INFO 4910,Benjamin Soltoff,bcs88@cornell.edu
+INFO 4940,Sharlane S. Cleare,ssc294@cornell.edu
+INFO 4997,Lawrence Edward Blume,lb19@cornell.edu
+INFO 4998,Walker Mcmillan White,wmw2@cornell.edu
+MATH 1006,Doug Knowles,ddk62@cornell.edu
+MATH 1011,Doug Knowles,ddk62@cornell.edu
```
NIDHI2023, line 173: Also these cases where a professor teaches multiple classes, maybe also make note of. Any unusual cases we could manually send an email, but it would help to have them all listed out

NIDHI2023 (commented):
Thanks for making such a speedy PR, it looks really good so far and the formatting for the csv seems to work out for most cases! I think other filter ideas could be setting the location to Ithaca instead of Cornell Tech, and some of the class levels could probably be lower since the 4000s tend to be research/project courses.
One other small note, in general when making branches, I'd suggest naming them like 'eric/<your task here>'

src/scripts/course-scrape.py
EricWeng23: Oops, I overlooked a mistake in this helper method. The naming of the definition was intended to be `getProfessorInfo()` .

src/classes.csv
EricWeng23: Screenshot 2025-02-09 at 10 36 33 PM

Hmm, this is an unusual case, thank you for pointing that out. I checked the JSON file for all the CHEM courses and it seems like the script is retrieving the professor's name correctly. The professor's last name as indicated for the course was indeed Chen, Prof. I think this is an issue with the Class Roster API, so it would be hard to check for this case which means we might have to check this manually...

src/scripts/course-scrape.py
```diff
@@ -0,0 +1,76 @@
+import requests
+import csv
+import json
+
+def grab_json(url):
+  return json.loads(requests.get(url).text)
+
+rosterURL = "https://classes.cornell.edu/api/2.0/config/rosters.json"
```
laylaliuuu, line 8: Looks good. Good data scraping.

laylaliuuu (approved):

src/classes.csv
EricWeng23: This has been addressed from the latest commit 
Screenshot 2025-02-18 at 12 02 39 AM

src/scripts/course-scrape.py
```diff
@@ -0,0 +1,129 @@
...
+        # Apoend the combined version of the courses
+        classes.append({
+          'Course': ', '.join(course_list),
+          'Professor': professor_name,
+          'Emails': professor_email
+        })
+else:
+    print("No professors teach multiple courses.")
+
+with open('../classes.csv', 'w') as file:
```
NIDHI2023, line 121: Small thing for organization - would it be possible to output the csv file within the scripts folder?

src/classes.csv
```diff
@@ -0,0 +1,108 @@
...
+MATH 2940,Frans Schalekamp,fms9@cornell.edu
+MATH 3040,Anil Nerode,an17@cornell.edu
+MATH 3110,Timothy R. Riley,trr22@cornell.edu
+MATH 3320,Dan Mihai Barbasch,dmb14@cornell.edu
+MATH 3340,Rachel M. Webb,rmw263@cornell.edu
+MATH 3360,Karola Meszaros,km626@cornell.edu
+ORIE 3120,Peter Frazier,pf98@cornell.edu
+ORIE 3310,David B. Shmoys,dbs10@cornell.edu
+ORIE 3741,Soroosh Shafiee,ss3789@cornell.edu
+PHYS 1102,Nam Jung Kim,nk652@cornell.edu
```
NIDHI2023, line 85: PHYS 1102 is apparently "Physics 1112 Supplement" so I think adding 'supplement' to the non-relevant words list would be a good idea. Also maybe "Honors" bc I think there were occasionally honors level cs courses

src/classes.csv
```diff
@@ -0,0 +1,108 @@
...
+"PHYS 3360, AEP 3630",Earl John Kirkland,ejk14@cornell.edu
+"STSCI 2100, STSCI 3200, BTRY 3020",Jeremy Entner,jfe35@cornell.edu
+"BTRY 3080, STSCI 3080",Florentina Bunea,fb238@cornell.edu
+"CHEM 3570, CHEM 1570",Cynthia Kinsland,clk10@cornell.edu
+"CS 1340, INFO 1260","Jon M. Kleinberg, Karen Levy","jmk6@cornell.edu, kl838@cornell.edu"
+"INFO 3152, CS 3152",Walker Mcmillan White,wmw2@cornell.edu
+"CS 3420, ECE 3140",Nils Napp,nen38@cornell.edu
+"ECON 3110, STSCI 2110, STSCI 3110",Kevin C Packard,kcp48@cornell.edu
+"INFO 3312, INFO 2951",Benjamin Soltoff,bcs88@cornell.edu
+"ORIE 3510, STSCI 3510",Jim Dai,jd694@cornell.edu
```
NIDHI2023, line 108: Great job with the multiple courses! And the list looks good overall 👍

src/scripts/course-scrape.py
```diff
@@ -0,0 +1,129 @@
...
+    field = ["Course", "Professor", "Emails"]
```
NIDHI2023, line 122: As we've talked about, we also need a 'has this course/professor used qmi before' field

NIDHI2023 (approved):
Logic + csv file looks good, great work finishing up this PR! Before merging, just add INFO/CS 4152 to the blacklisted courses and the word "Thesis" to irrelevant_courses, I think that'd be useful for one last filtering edit

PR #924, opened 2025-02-09
Join the Queue UI Updates

anchen9:
### Summary 

Implement Lucy's [redesign](https://www.figma.com/design/WFhJLWyi7EbCP08jKrTXxc/Lucy-FA24-Working-File?node-id=5-8921&t=McFSLWMx1FW8vpq4-0) for the Join the Queue section including adjusting padding and adding markings to denote required fields. Implemented Error State. Add Files in progress. 

Previous:
Screenshot 2025-02-23 at 1 55 36 PM

Updated:
Screenshot 2025-03-02 at 7 19 03 PM
Screenshot 2025-03-02 at 7 20 04 PM
Screenshot 2025-03-02 at 7 20 22 PM

Error State:
Screenshot 2025-03-10 at 11 48 13 AM

- Adjusted text sizes
- Adjusted placeholder text
- Edited padding/spacing
- Added *Required to the top right and asteriks to every field
- Added Error State
- Added FadeIn/FadeOut animations for the error state
- Made styling consistent for all fields
- Began the portion for adding files (adding the svg for the button and making the container
- Commented out file part for merge purposes

### Test Plan 

- inspect element to see if adjustments were made
- tested submitting when different fields were empty.

### Notes 

- the error state does not span the full width of the join the queue box, asked Renee & confirmed to be okay. 
- implemented horizontal line between location and question which is not on Figma, but suggested by Renee
- add file feature for the future?

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


NIDHI2023 (commented):
Great work so far! One note - I don't think the file part is required, so it doesn't need the asterisk. Also once you're done working on this PR (ik it's still in progress), make sure you attach screenshots of the before vs. after and the figma designs for future reference

src/components/includes/AddQuestion.tsx
```diff
@@ -217,32 +219,37 @@ const AddQuestion = ({ course, session, mobileBreakpoint, showProfessorStudentVi
...
-                                                    isPrimary={true}
-                                                    select={true}
-                                                />
-                                            ))}
+                                <div className="topRow">
+                                    <div className="disclaimerContainer">
+                                        <p className="text"><span className="required"> * </span>Required</p>
+                                    </div>
+                                    <div className="tagsMiniContainer">
+                                        <p className="header">Select a Category<span className="required"> * </span></p>
```
EricWeng23, line 227: Overall the code is good. An optimization I would make is probably factor out the asterisk into a reusable component like `const Asterisk = () => <span className="required"> * </span>;`. This way we don't have to repetitively nest span tags inside another jsx element.

anchen9: Ooh thanks for catching that! I just added that in!

NIDHI2023 (changes requested):
Good work with the sections and the padding, it looked good on my end. I do think having the error state span the whole width and having the red background kind of fade in a little slower might also look better.

Also just my opinion but you can def check with Lucy too, I think it's fine if the error state appears and disappears when you type/clear a question, but I think it should be consistent for the selections as well. I noticed it didn't do that with assignment/tags

NIDHI2023 (approved):
thanks for making the changes!

PR #926, opened 2025-02-22
Adding ability for non-Cornell users to login

NIDHI2023:
### Summary 
This PR changes the authorization for logging in to provide the option to sign in as a Cornell or non-Cornell user.
image

### Test Plan 

Tested to see if you can log in with different users

### Notes 

Hopefully this makes testing less of a pain...
Currently this option isn't in production and mainly for us developing/testing, but if it'd be useful we can push it to production.
### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/pages/LoginView.tsx
+++ b/src/components/pages/LoginView.tsx
@@ -35,7 +35,7 @@ const LoginView: React.FC = () => {
             authProvider.setCustomParameters({
                 hd: 'cornell.edu',
             });
-        } else if (process.env.REACT_APP_IS_STAGING !== 'true') {
+        } else if (process.env.NODE_ENV !== 'production' && process.env.REACT_APP_IS_STAGING !== 'true') {
             authProvider.setCustomParameters({
                 prompt: 'select_account'
             });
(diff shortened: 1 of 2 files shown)


src/components/pages/LoginView.tsx
```diff
@@ -35,6 +35,10 @@ const LoginView: React.FC = () => {
             authProvider.setCustomParameters({
                 hd: 'cornell.edu',
             });
+        } else if (process.env.REACT_APP_IS_STAGING !== 'true') {
+            authProvider.setCustomParameters({
```
laylaliuuu, line 39: Is it necessary to add a else{} to avoid any cases besides If{} and else if{} to be safe? Like, are there any edge cases?

laylaliuuu (commented):
I tried to log in with a non-Cornell user email, and it worked! Good job!!

src/components/pages/LoginView.tsx
NIDHI2023: I think considering that the original code didn't consider any other cases, there shouldn't be edge cases but this is a good thought. I'll add an explicit check in the if to make sure it's not production

laylaliuuu (approved):
It worked!

PR #928, opened 2025-03-03
Layla/finish edit classes frontend

laylaliuuu:
### Summary 
Finishes Editing classes frontend, resolved Sophie's PR from last semester. 

1. Fix the inconsistency in hover effects in former classes 
2. Cleaned up the style code in .tsx file to .cscc

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/styles/courses/CourseCard.scss
```diff
@@ -1,65 +1,156 @@
...
+    border: 2px solid #D8D8D8;
+    overflow-wrap: anywhere;
+
+    
+    selectedBackgroundColor: "#F5F5F5";  // the default background color
+    selectedBorderColor: "#D8D8D8";  // the default border color
+    role-color-ta:"#BF7913"; // the color for "TA" 
+    role-color-professor: "green"; // a purple thats closer to our brand colors- #726CFF
+    selectedBackgroundColor-role-student: "rgba(214, 234, 254, 0.4)"; //  "Student" role background color
+    selectedBorderColor-role-student: "#77BBFA"; //  "Student" role border color
```
NIDHI2023, line 21: I think the variable names in the css files are a nice addition to make the code more organized

NIDHI2023 (changes requested):
Great work, the hover issue seems to be fixed for me! A few more things, I think Richard mentioned a change to the font style specifically, I think the font-size Sophie took care of already.  
image

Also, when running I was getting this new warning that should probably be addressed:
image

You could resolve these by adding 'filterOnActiveAndRole' and 'currentCourses' into the useEffect dependency arrays as suggested. However, since filterOnActiveAndRole is an inline function, React sees it as a "new" function on every render of the CourseSelection component. So even if the logic inside filterOnActiveAndRole does not change, its reference will still be different after each render, which could unnecessarily trigger a lot of useEffect calls.

I think you could look into useCallback to ensure that the function reference remains stable unless one of its dependencies changes - so something like this:
`const filterOnActiveAndRole = useCallback(() => {
    return allCourses
    ....
        });
}, [allCourses, CURRENT_SEMESTER, user.userId]); `

src/styles/courses/CourseCard.scss
anchen9: I also like the variable names as well, I think its a great idea! I'm not sure if its an error on my end, but for some reason in the CourseCard.scss file the variables are not recognized as variables with the $ starting character?

anchen9 (commented):
Hi Layla! The hover state and documentation look great, good work on finish up Sophie's PR! Could you attach before & after screenshots to the PR as well?

src/styles/courses/CourseCard.scss
NIDHI2023: Good catch Annie, this is what it looks like on my end with and without the $:
image

src/styles/courses/CourseCard.scss
```diff
@@ -1,65 +1,156 @@
...
+            align-items: center;
+            height: 32px;
+            padding: 10px 17px;
+            text-align: center;
+            gap: 10px;
+            border-radius: 100px;
+
+            //check coursecard for hex color of prof
+            border: 2px solid #BF7913;
+            color: #BF7913;
```
NIDHI2023, line 72: I think the Prof color is hardcoded here, maybe use one of the variables you defined?

src/components/includes/CourseSelection.tsx
```diff
@@ -32,14 +32,30 @@ function CourseSelection({ user, isEdit, allCourses }: Props): React.ReactElemen
...
-        setCurrentCourses(allCourses.filter((course) => {
-            return course.semester === CURRENT_SEMESTER;
-        }));
+        setCurrentlyEnrolledCourseIds(new Set(user.courses));
+    }, [user.courses]);
 
-        setFormerCourses(allCourses.filter((course) => {
-            return course.semester !== CURRENT_SEMESTER;
-        }));
+    const filterOnActiveAndRole = () => {
```
EricWeng23, line 41: I think an optimization that can be made for the `filterOnActiveAndRole` function is to use the React Hook `useMemo` to "memoize" the calculation of all the filtering and sorting of all courses. The current implementation would have to run the filtering and sorting every time the page renders, which may result in latency. This is probably not a significant issue right now since QMI only has a limited number of courses, but once the course list expands for `allCourses`, it would be computation-heavy. An example of `useMemo` would be:

`const filterOnActiveAndRole = useMemo(() => {
 ...
}, [dependency array])`

src/components/includes/CourseCard.tsx
```diff
@@ -11,53 +12,101 @@ type Props = {
...
+            }
         >
+            <Grid container direction="row" justifyContent="space-between" style={{ height: "58px" }}>
+                {roleString ? (
+                    <Grid container item className="courseColor" xs={6}>
+                        <span
+                            className="courseRole"
+                            style={{
+                                border: "2px solid " + roleColor,
+                                color: roleColor,
```
NIDHI2023, line 73: So here's where Sophie's trying to use the roleColor variable above - if the css variables don't work out you might just need to hardcode values like "green" for professor and "#BF7913" for TA

src/components/includes/CourseCard.tsx
```diff
@@ -11,53 +12,101 @@ type Props = {
...
     }
+
     return (
         <div
-            className={`CourseCard ${selected && editable ? 'selected' : ''} ${
-                inactive ? 'inactive' : 'active'
-            }`}
+            className={`CourseCard ${selected && editable ? "selected" : ""} ${inactive ? "inactive" : "active"} 
+            ${roleString === "" ? "editable" : "ineditable"}`}
             onClick={selectCourse}
```
laylaliuuu, line 59: I don't think I need this, as all of them should be ineditable (clickable?)

src/components/includes/CourseCard.tsx
```diff
@@ -11,53 +12,101 @@ type Props = {
...
+    if (role === "ta") {
+        roleString = "TA";
+        roleColor = "#var(role-color-ta)";
+    } else if (role === "professor") {
+        roleString = "PROF";
+        roleColor = "#var(role-color-professor)";
+    } else {
+        selectedBackgroundColor = "selectedBackgroundColor-role-student";
+        selectedBorderColor = "selectedBorderColor-role-student";
     }
```
laylaliuuu, line 53: I feel like this can be deleted too, but we can talk about it tomorrow, because there are few places that used roleString and i don't know if that will change the behavior of the roles.

src/styles/courses/CourseCard.scss
laylaliuuu: should i just delete it here? Because I dont think it make sense here.

src/styles/courses/CourseSelection.scss
```diff
@@ -10,42 +27,114 @@
...
+                border: 2px solid #D8D8D8;
+                border-radius: 15px;
+                margin-top: 8px;
+                margin-right: 72px;
+
+                input {
+                    flex: 1;
+                    border-radius: 15px;
+                    border: none;
+                    font-size: 18px;
```
laylaliuuu, line 84: I added the font here, but I think "Roboto" is already coded as the font for the entire page at top, but I just added it here again. Also I changed the font size to a little bit smaller to match the design.

NIDHI2023 (approved):
Looks great :) Thanks for all your work!

PR #929, opened 2025-03-03
Automate professor email outreach

NIDHI2023:
### Summary 

Building on[ this PR](https://github.com/cornell-dti/office-hours/pull/923) creating the information csv file, this PR enables a user to run a script to send out emails to those professors. It has a slightly different message depending on if the professor, or course has used QMI before, or if they're completely new.

- Both scripts requires user to have python and possibly install some packages with pip

### Test Plan 

Some sample output from the script:

image

This was also used to send out some emails this semester.

### Notes 

Before running this script, a human should manually check the csv list as a sanity check. There are some things that are hard to code and might need to be sent manually (ex: a professor with the last name van Dover, but the script would address it to Professor Dover). Also would be helpful to just make sure there are no accidental repeat classes in there (ex: CS 4300/INFO 4300).

These email PRs handle a major part of automation, which was one of QMI's main goals for this semester. The last part would be to automate course creation using some old PRs and possibly finding a way to automate updating constants.ts.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/scripts/outreach_email.py
```diff
@@ -0,0 +1,113 @@
...
+                profemails = [entry.strip() for entry in profemails.split(',')]
+            else:
+                profemails = [profemails]
+            # Handle multiple courses
+            course_final = course
+            if ',' in course:
+                course_final = [c.strip() for c in course.split(',')]
+                course_final = "/".join(course_final)
+            
+            # used value of 0 means only the course has used QMI. value of 1 means the professor has used QMI. value of 2 means both the professor and course used QMI.
```
anchen9, line 76: Hi Nidhi! Thanks for all of your hard work on automating the professor email outreach! I think your handling of different cases, readability of your code, and code organization are great! I like that you wrote comments on what used values indicate, I wonder if it might improve readability and maintainability if this was separated as constants or an enum?

EricWeng23 (approved):

PR #931, opened 2025-03-03
Eric / Implement Firebase Function to Monitor Question Status Change

EricWeng23:
### Summary 

*UPDATE 2:*
Refactored the firebase cloud function `onQuestionStatusUpdate`. Instead of maintaining a `resolvedQuestionArray` field inside of the session documents and having to update the session everytime after calling `removeQuestionDisplayFeedback` to remove the processed questions, maintain a `recentlyResolvedQuestion` field inside of specific user document.

The previous logic of updating session document would require modifying firebase security rules because we wouldn't want non-admin users to update office hour sessions without restrictions, so only Professors or TA should have update/write permission. Thus, the approach of updating the user document after the questions are resolved and processed does not require changing any of the existing firebase security rules. Doing so also prevents the feedback modal broadcasting to other users which was an existing bug prior to this change.

*Note:* 
- Only a handful of Snapshot is called
- Feedback Modal doesn't get broadcasted to other users

https://github.com/user-attachments/assets/7b560b0b-266b-4287-ab15-c126f53da3fd

This is a follow-up PR to alleviate the overactive listener that was spiking up Firebase cost. This PR addressed this issue by  implementing a new firebase function `onQuestionStatusUpdate` which compares the status of the new and updated question. If status changed to resolved, then we can call `removeQuestionDisplayFeedback`. This allows developers to monitor question changes on the server-side rather than the client-side. Specifics is in Richard's past PR: https://github.com/cornell-dti/office-hours/pull/888

The number of `onSnapShot` calls reduces

*Note*: Currently this implementation tries to update the reference to a session document with an array of recentlyResolvedQuestions, but this may not be an optimal and permanent solution.

### Test Plan 
https://github.com/user-attachments/assets/b68e0486-6c25-4951-89bf-438e9c189d44

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -506,9 +506,8 @@ exports.onQuestionStatusUpdate = functions.firestore
             // Update the resolvedQuestionsArray field in the session document if it exists
             return userDoc.update(
                 {
-                    // An array that keeps track of questions that were just resolved by a course staff
-                    // Each question would be removed from the array once its feedback modal has been processed
-                    // Consists of objects with questionId, askerId, and resolvedAt fields
+                    // Keeps track of the most recent question that was resolved
+                    // Object with questionId, askerId, and resolvedAt fields
                     // questionId: the id of the question that was resolved
                     // askerId: the id of the user who asked the question
                     // resolvedAt: the time the question was resolved
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -508,6 +508,7 @@ exports.onQuestionStatusUpdate = functions.firestore
                 {
                     resolvedQuestionsArray: admin.firestore.FieldValue.arrayUnion({
                         questionId,
+                        sessionId,
                         resolvedAt: admin.firestore.Timestamp.now(),
                     })
                 },
@@ -518,6 +519,7 @@ exports.onQuestionStatusUpdate = functions.firestore
                         resolvedQuestionsArray: [
                             {
                                 questionId,
+                                sessionId,
                                 resolvedAt: admin.firestore.Timestamp.now(),
                             },
                         ],
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -313,6 +313,10 @@ questionStatusNumbers.set("no-show", [0, 0, 0]);
 exports.onQuestionUpdate = functions.firestore
     .document('questions/{questionId}')
     .onUpdate(async (change) => {
+        // Log when the function is triggered
+        functions.logger.info("Function triggered", {
+            questionId: change.after.id,
+        });
         // retrieve old and new questions
         const newQuestion: FireQuestion = change.after.data() as FireQuestion;
         const prevQuestion: FireQuestion = change.before.data() as FireQuestion;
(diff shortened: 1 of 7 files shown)


functions/src/index.ts
```diff
@@ -484,4 +488,33 @@ exports.onQuestionUpdate = functions.firestore
...
+
+        // Derive session ID
+        const sessionId = newQuestion.sessionId;
+
+        // Derive statuses
+        const newStatus = newQuestion.status;
+        const prevStatus = prevQuestion.status;
+
+        // Only proceed when newQuestion status is resolved
+        if (prevStatus !== "resolved" && newStatus == "resolved") {
```
NIDHI2023, line 508: Use "===" instead of "==" here

NIDHI2023:
renaming title so WIP check can pass

functions/src/index.ts
```diff
@@ -484,4 +488,33 @@ exports.onQuestionUpdate = functions.firestore
             totalWaitTime: admin.firestore.FieldValue.increment(waitTimeChange),
             totalResolveTime: admin.firestore.FieldValue.increment(resolveTimeChange),
         });
-    });
\ No newline at end of file
+    });
+
+exports.onQuestionStatusUpdate = functions.firestore
```
NIDHI2023, line 493: The logic looks good, and I think it's a cleaner solution to maintain a "questions resolved" array for each session rather than querying all the questions. I think this array also shouldn't get unreasonably large to store.

NIDHI2023:
I see that you were able to deploy this new function to QMI-test, which is great! We'll have to do this for qmi-prod when it gets merged into release too, so I would add a note about that.
image

functions/src/index.ts
```diff
@@ -484,4 +488,49 @@ exports.onQuestionUpdate = functions.firestore
...
+                    })
+                },
+            ).catch(() => {
+                // If the resolvedQuestionsArray field does not exist, create it
+                return sessionDoc.set(
+                    {
+                        resolvedQuestionsArray: [
+                            {
+                                questionId,
+                                sessionId,
```
NIDHI2023, line 523: this field of sessionId might be a bit redundant, because to get this array in the first place we have to do something like sessionCollection.doc(sessionId).resolvedQuestionsArray, so we'd already know this whenever we're getting this information? and I think all the items in the array would have the same sessionId anyways

functions/src/index.ts
```diff
@@ -484,4 +488,49 @@ exports.onQuestionUpdate = functions.firestore
...
+        if (prevQuestion.status !== "resolved" && newQuestion.status === "resolved") {
+            const sessionId = newQuestion.sessionId;
+
+            // Retrieve the session document reference 
+            const sessionDoc = db.doc(`sessions/${sessionId}`);
+
+            // Update the resolvedQuestionsArray field in the session document if it exists
+            return sessionDoc.update(
+                {
+                    resolvedQuestionsArray: admin.firestore.FieldValue.arrayUnion({
```
NIDHI2023, line 509: If I'm understanding correctly, this array is basically keeping track of questions that are resolved by TA but haven't been shown feedback yet? Once they're shown feedback they're removed from the array, so once a session is over the less entries we have the better? I would definitely add some documentation to describe what this field is for

functions/src/index.ts
EricWeng23: I agree, `sessionId` is indeed a redundant field because in `SessionView` there is a session prop passed in, so `sessionId` is easily retrievable.

NIDHI2023 (approved):
The functionality works for me. Great thinking with updating the users collection so we don't have to make large changes to the security rules!

PR #932, opened 2025-03-16
Upgrade cloud functions

NIDHI2023:
### Summary 

Upgrading Firebase Functions to v6, Firebase to v11.
- Also upgraded rxfire to v6 and rxjs to v7

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/components/includes/TopBar.tsx
```diff
@@ -59,21 +75,33 @@ const TopBar = (props: Props) => {
...
+                    if (Notification.permission === "granted") {
+                        addNotification({
+                            title: notif.title,
+                            subtitle: notif.subtitle,
+                            message: notif.message,
+                            native: true,
+                        });
+                    }
+
+                    // addNotification({
```
burninglilies, line 89: dont forget to remove comment here, new check is good for notifications

src/components/includes/TopBarNotifications.tsx
```diff
@@ -55,7 +55,10 @@ const TopBarNotifications = (
 
     useEffect(() => {
         if(notificationTracker !== undefined && !hasViewed && dropped) {
```
burninglilies, line 57: if (notificationTracker && !hasViewed && dropped) {
  (async () => {
    await periodicClearNotifications(user, notificationTracker);
  })();
}
I suggest immediately calling the async function instead of defining it and then calling it separately

burninglilies (approved):
CRAZYY WORK Nidhi, 42 files changed!! I did a naive test of most functions and monitored console logging and so far and so far so good. A couple things from console I was able to note were DOM warnings, invalid document warnings from creating tags/assignments, and non unique "key" props from csv exports (I can show you when we meet after spring break). I didn't test too seriously for edge cases but despite the warnings it seems to work fine for me locally? Also made a few styling suggestions but I think it should be good to push onto master and see if any issues pop up for the team for extra testing before we try and push it into release.

functions/src/index.ts
```diff
@@ -1,4 +1,5 @@
-import * as functions from 'firebase-functions';
+// eslint-disable-next-line import/no-unresolved
```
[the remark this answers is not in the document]
burninglilies: I commented this at first but noticed you add these throughout the code I assume intentionally so you can disregard this

PR #933, opened 2025-03-23
Eric/Fixed Bug where Question Feedback Prompt is not displayed

EricWeng23:
### Summary 

While implementing the function for this <a href="https://www.notion.so/1740ad723ce1817bb989e7d8329019ea?v=1740ad723ce181809fa7000c727ccc7e&p=1b90ad723ce180fea0eff8412b4004b9&pm=s"> ticket </a>, I realized a bug where although a prof/TA resolved a question, there was no question feedback prompts shown on the student side. I checked the firebase function loggers as shown below:
image
The user document was not successfully updated due to the invalid field `recentlyResolvedQuestion.resolvedAt`. This field is not necessary for the functionality of monitoring questions, and was intended to be used for debugging. This PR provides remedy for this issue by simply removing `resolvedAt` entirely. 

The updated version of the cloud function `onQuestionStatusUpdate` is already deployed to the test QMI firebase

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -510,11 +510,9 @@ exports.onQuestionStatusUpdate = functions.firestore
                     // Object with questionId, askerId, and resolvedAt fields
                     // questionId: the id of the question that was resolved
                     // askerId: the id of the user who asked the question
-                    // resolvedAt: the time the question was resolved
                     recentlyResolvedQuestion: {
                         questionId,
                         askerId: userId,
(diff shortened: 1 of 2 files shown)


NIDHI2023 (approved):
Tested - functionality seems to still work.
Edit: the feedback modal not working might also have been because I was trying out deployments that weren't working lol but also good catch with the logs!

PR #934, opened 2025-03-23
Eric/Calculating Student Per TA Statistic for TA Block Component

EricWeng23:
### Summary 

*Revision 3*
Modifications were made to this PR by simplifying the logic when updating session document. Previously this PR used `serverTimestamp` for a consistent synchronized clock, but that'll cause an infinite loop of the cloud function ` onStudentJoinSession` because to get the timestamp it would require writing to an existing document which will repeatedly trigger the cloud fn. This PR avoids this issue by discarding the usage of server timestamps and essentially just set ratio to the number of TAs available. Note that besides the `studentPerTaRatio`, there 's now an `hasUnresolvedQuestion` field for type `FireBaseSession`. Here are the diffferent scenarios

Scenario 1: Session has not started
- `hasUnresolvedQuestion` is `false` to indicate that there are no questions present in the session
- `studentPerTaRatio` would be set the number of TAs assigned

Scenario 2: Session has started
- `hasUnresolvedQuestion` is `true`
- `studentPerTaRatio` is computed

There were also modifications made to `ProfessorOHInfo`, particularly `mutateSessionOrSeries` which is a function responsible for creating and editing sessions. Previously whenever the session is updated, this function will reset the metrics (e.g totalQuestions, assigedQuesions) to 0 which will mess with the ratio computation logic. Suppose a professor decided to assign another TA to an ongoing session, the ratio would be inaccurate as the data are all resetted. Furthermore, for consistency I initialized the fields `studentPerTaRatio` and `hasUnresolvedQuestion` when a session is created in `mutateSessionorSeries`

*When session is created without TAs assigned*
https://github.com/user-attachments/assets/11b0bb8b-b267-43ac-9d12-199c46f22479

*When session is created with some number of TAs assigned*
https://github.com/user-attachments/assets/824ec734-6844-435c-ba21-1bcb914c9c42

*Note how when TA is added, the metrics don't change*
https://github.com/user-attachments/assets/0d6b7d41-554d-4818-960b-95e407bbccb3

*Revision 2*
Modifications were made to this PR to address comments that include:
- Modified firebase security rules to loosen the restriction when updating session documents.
- Synchronized clock time (avoid using client local machine time, use firebase serverTimestamp instead)
- Added query indexes when querying for composite question documents 
- Refactored `fetchNumberOfStudentsPerTA` into a cloud function `onStudentJoinSession` to allow updating the new fields `serverTimestamp` and `studentPerTaRatio` on the server side triggered session document changes rather than having to update the field on the client side whenever a question is assigned.

To implement a synchronized clock using `admin.firestore.FieldValue.serverTimestamp`, it requires using `set()` or `update()` to include a server-generated timestamp in the written data. Therefore I modified the security rules like such:
```
function fieldsCanBeWritten(fields) {
    	return request.resource.data.diff(resource.data).affectedKeys().hasOnly(fields);
   }
match /sessions/{session} {
      	allow read: if request.auth.token != null;
        //allow create, update: if request.auth.token!= null;
        allow create, update: if sessionOrSeriesCanBeWritten() || fieldsCanBeWritten(["serverTimeStamp", "studentPerTaRatio"]); 
```
This way we can still allow updates to be made to session documents when the only affected fields are in this case `serverTimestamp` and `studentPerTaRatio`

Added query indexes:
image

Testing via local:

https://github.com/user-attachments/assets/13fb2ff1-1f78-4940-abbc-6bbde8eee779

When a question is assigned, removed, or resolved it will recalculate the ratio and update it in the corresponding session document. 

*Note*: Also tried changed my local machine time to be 5 min ish behind, and functionality works the same as the vid.

*Revision 1*
This PR implemented the functionality of this <a href="https://www.notion.so/1740ad723ce1817bb989e7d8329019ea?v=1740ad723ce181809fa7000c727ccc7e&p=1b90ad723ce180fea0eff8412b4004b9&pm=s"> ticket </a>. For the impending new design consisting of the TA Block, we need statistics of how many students per TA during that OH session.  

By convention, I chose to start tracking the number of students per TA once the session has BEGUN (meaning that current time > session start time). If the session has not started yet, then `getNumberOfStudentsPerTA` will by default return the number of TAs in that session. Note that if there are no TAs assigned to that session (although rarely happens), it will return 0 to avoid zero division. 

We will only track the questions with `status` of `assigned`, because that indicates that the question is being handled by a particular TA.  Although the `user` collection in firebase has the fields `assignedQuestions` and `totalQuestions`, these metrics do not constitute the UNIQUE number of students in that OH session since the same student may ask multiple questions. 

Thus we will ensure that the `askerId` of all the assigned questions during that OH session are unique using a set data structure to record these ids.

### Test Plan 
Testing via local server

https://github.com/user-attachments/assets/64f1628a-94c6-4f4a-bf05-c315bb9aafd2

Note: There are 2 TAs assigned that the session above

When there are no assigned questions, return 0.
When a question is resolved, update the value with the new collection of assigned questions

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


anchen9 (commented):
Hi Eric! Great work on the students per ta calculations! I liked that your function was well documented & easy to follow! I also think returning 0 early to handle potential division by 0 errors is a smart preventative measure! Is the useEffect in SessionView.tsx fetching the number of students per ta supposed to be commented out?

src/firebasefunctions/session.ts
```diff
@@ -63,4 +63,61 @@ export const deleteTaAnnouncement = (
...
+ * @param session - A FireSession object that represents the session.
+ * @returns A number that represents the ratio of students per TA.
+ */
+export const getNumberOfStudentsPerTA = async (
+    session: FireSession
+) => {
+    // Get the start time of the session
+    const startTime = session.startTime.toDate();
+    // Get the current time
+    const now = new Date();
```
NIDHI2023, line 85: If the client's clock is off (for example mine is 5 minutes behind) this could cause issues right before the sessions actually starts. Maybe you could use admin.firestore.FieldValue.serverTimestamp() to use the server times to standardize across machines?

src/firebasefunctions/session.ts
```diff
@@ -63,4 +63,61 @@ export const deleteTaAnnouncement = (
...
+    // Get the number of TAs in the session
+    const numberOfTAs = session.tas.length;
+    // If there are no TAs return 0 to avoid division by 0
+    if (numberOfTAs === 0) {
+        return 0;
```
NIDHI2023, line 90: I think it could be useful on the frontend to know the difference between when there are 0 students in the queue (which would also have a 0 value) vs no TAs assigned. I think this could be a different sentinel value like -1 which signifies we should display something like "No TAs assigned"?

src/firebasefunctions/session.ts
NIDHI2023, line 85: If the client's clock is off (for example mine is 5 minutes behind) this could cause issues right before the sessions actually starts. Maybe you could use admin.firestore.FieldValue.serverTimestamp() to use the server times to standardize across machines?

src/firebasefunctions/session.ts
```diff
@@ -63,4 +63,61 @@ export const deleteTaAnnouncement = (
...
+        return numberOfTAs;
+    }
+    // Get the session ID
+    const sessionId = session.sessionId;
+    // Get the reference to the question collection but only containing questions 
+    // from this specific session and are assigned to a TA
+    const questionsRef = firestore
+        .collection("questions")
+        .where("sessionId", "==", sessionId)
+        .where("status", "==", "assigned");
```
NIDHI2023, line 104: Not 100% sure on this but I think when you create a new query for the first time you have to add an index to the query rules in firestore. Maybe log for errors here just in case and test once more
image

src/firebasefunctions/session.ts
```diff
@@ -63,4 +63,61 @@ export const deleteTaAnnouncement = (
...
+    
+    // Create a set to store unique students (same student may ask multiple questions)
+    // We are using number of UNIQUE students as metric rather than just the total number of 
+    // questions without accounting for their askers
+    const uniqueStudents = new Set<string>();
```
NIDHI2023, line 109: Good idea using a Set 👍

src/firebasefunctions/session.ts
```diff
@@ -63,4 +63,61 @@ export const deleteTaAnnouncement = (
...
+
+    const snapshot = await questionsRef.get();
+    if (snapshot.empty) {
+        return 0;
+    }
+    // Iterate through the question documents and add the UNIQUE student ID to the set
+    snapshot.forEach((doc) => {
```
NIDHI2023, line 116: Looking at this stackOverflow comment, I think just adding to the set (since we're adding strings which should be easily comparable, not objects) is fine rather than doing this extra check:
https://stackoverflow.com/questions/69732773/duplicate-checking-vs-creating-a-set-typescript

NIDHI2023 (changes requested):

NIDHI2023:
Nice work with this! Your logic looks good for the most part and you considered a lot of the edge cases. Thanks also for the great PR + code documentation

NIDHI2023 (changes requested):
I tested adding multiple students to the queue and the ratio gets updated correctly. Adding a question to a session with no TA also correctly returns a signal value of -1.  Great work! It runs really smoothly. I just have a few suggestions for some edge cases, and then it will be good to merge:

As you mentioned when TAs are added during a session, the totalResolved questions gets set to 0. However, I think we should fix it to not reset because that could become a potential bug in the future.

I also noticed that when you create a session with no TA initially, there is no student/ta field - would it make sense to initialize it to -1? I added a TA and saw that the ratio field was created and set to 0. Then when I deleted the TA again it was set to -1, so I feel like we should have consistency.

NIDHI2023:
To look at the totalQuestions field, check out `mutateSessionOrSeries` in ProfessorOHInfo

NIDHI2023 (approved):
Great job with the fixes for totalQuestions and the other fields! I think this will be a better design choice going forwards. Tested a bunch and everything seems to work good 😄

PR #935, opened 2025-03-23
TA Dashboard - Preparation & Student View Tabs (In Progress)

anchen9:
### Summary 

Implementing the TA Dashboard frontend designs from this figma [file](https://www.figma.com/design/gKqRwgffXSKNB5PYQlXHS0/Sia-FA'24-Working-File-(%2B-Renee-SP25-Work)?node-id=1-124&p=f&m=dev), specifically the Preparation and Student View tabs. Setting up the TA Dashboard routing and layout.  

Working on these two tickets: [preparation](https://www.notion.so/TA-Dashboard-Frontend-Preparation-Tab-19b0ad723ce1805298e4d3c68ba61011?pvs=4), [student view](https://www.notion.so/TA-Dashboard-Frontend-Student-View-Tab-1b90ad723ce18018af3ed418f1eb9afc?pvs=4)

Current Appearance: 
Screenshot 2025-03-22 at 9 33 44 PM

Screenshot 2025-03-22 at 9 33 17 PM

- Reused code from Professor View like Student View, Sidebar to implement TASidebar, TAView, TAStudentToggle.
- Created scss files for new pages/components and added css files to index.scss
- Added routing for the preparation and student view tabs in App.tsx
- Adjusted toggle styling to have even width & blue text color when selected

### Test Plan 

- I tried clicking each tab to make sure the routing worked
- try accessing the dashboard as a prof and as a student. 

### Notes 

IN PROGRESS items:
- student query trends component
- resources component
- styling fixes to align colors to brand kit

Note: reused code from ProfessorStudentView, will keep an eye on whether this causes any issues with TA permissions.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


NIDHI2023 (changes requested):
Great work setting this up so far! I see that the updated styling for the "Queue" and "Dashboard" components has been added.
When I was looking at the mobile view, I think disabling scrolling and adding styling to make content fit smaller widths might be a better option.
image
One other small suggestion is to have the cursor change on hover so there's some indication the user can click.

src/components/pages/TAStudentView.tsx
```diff
@@ -0,0 +1,234 @@
+import React, { useEffect, useState } from "react";
```
NIDHI2023, line 1: I know you mentioned making this file for the routing above, but I would also test if using the component ProfessorStudentView works for TAs, and if so, refactor that file to just be like "AdminStudentView" and modify logic to vary based on TA or student. I'm thinking it shouldn't be too different, since both Professors and TAs are simply acting as a student in this view, so that will help with having a cleaner codebase with less repetition.

NIDHI2023 (commented):
Also in both the Preparation + Student View tabs, the dropdown gets cut off:
image

src/components/includes/TASidebar.tsx
```diff
@@ -0,0 +1,69 @@
...
+                                <div className="lineGraphIconContainer">
+                                    <LineGraph/>
+                                </div>                                
+                                Analytics
+                            </button>
+                        </Link> */}
+                        <Link to={"/ta-student-view/course/" + courseId}>
+                            <button type="button" className={css(selected === "student")}>
+                                <div className="iconContainer">
+                                    <ViewIcon />
```
NikhillA, line 57: This look great to me! I can clearly understand your changes and what specific impacts they have on the UI. Is it possible to reduce redundancy in this section of code here? I only ask as I noticed some lines being repeated. Maybe each individual button can be represented as a generic button with specialization.

NikhillA, line 57: This look great to me! I can clearly understand your changes and what specific impacts they have on the UI. Is it possible to reduce redundancy in this section of code here? I only ask as I noticed some lines being repeated. Maybe each individual button can be represented as a generic button with specialization.

PR #941, opened 2025-04-14
Eric/ OH Busyness Regression Model (in progress)

EricWeng23:
### Summary 

This PR implements the basics of the model by loading and preprocessing the data using Bag of Words representation to capture the keywords of the question contents before feeding it into LSTM (Long Short-Term) model 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/scripts/busyness/dummy_data.csv
+++ b/src/scripts/busyness/dummy_data.csv
@@ -1,0 +1,51 @@
+questionId,content,hasPrimaryTag,hasSecondaryTag,wasNotified,waitTimeBeforeAssignment,responseTimeAfterAssignment,totalDuration
+SZFjlyOOGqgKPOrhICjP,"How do I implement a FIFO queue in OCaml?",0,0,1,00:26:01,00:00:04,00:26:06
+SZFjlyOOGqgKPOrhICjP2,"Can someone explain what a functor is in OCaml and how to use it?",1,1,1,00:12:30,00:05:19,00:17:49
+SZFjlyOOGqgKPOrhICjP3,"What's the difference between inheritance and composition in OOP?",0,1,0,00:03:15,00:08:30,00:11:45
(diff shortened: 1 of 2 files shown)


src/scripts/busyness/dummy_data.csv
```diff
@@ -0,0 +1,51 @@
+questionId,content,hasPrimaryTag,hasSecondaryTag,wasNotified,waitTimeBeforeAssignment,responseTimeAfterAssignment,totalDuration
```
NIDHI2023, line 1: Could another feature be whether the professor has enabled a timer for that class or not? Those tend to be the large classes that have higher wait times

src/scripts/busyness/dummy_data.csv
```diff
@@ -0,0 +1,51 @@
...
+SZFjlyOOGqgKPOrhICjP,"How do I implement a FIFO queue in OCaml?",0,0,1,00:26:01,00:00:04,00:26:06
+SZFjlyOOGqgKPOrhICjP2,"Can someone explain what a functor is in OCaml and how to use it?",1,1,1,00:12:30,00:05:19,00:17:49
+SZFjlyOOGqgKPOrhICjP3,"What's the difference between inheritance and composition in OOP?",0,1,0,00:03:15,00:08:30,00:11:45
```
NIDHI2023, line 4: I'm also wondering if it would make sense to keep track of timeAsked, course, and day of week. For example if you knew you were looking at CS 1110 you'd expect higher wait times than CS 3780. And maybe there's a pattern of CS 1110 on Wednesdays that are busier (when assignments are due, but also possible those days are variable). While these patterns might be stronger than the content of the question, it won't have as much as data however.

But the end goal is to have an estimate for different days of the week, for different 15 or 30min blocks of time (we can probably change this depending on what works better), for each course.

PR #950, opened 2025-04-20
Nikhill/student feedback

NikhillA:
### Summary

This PR adds to the firebase by storing three additional ratings for TA's in the 'users' collection on firebase so that each rating is tied to each individual TA. 

 The three new ratings are Organization, Efficiency, and Overall experience and are collected through three new rows of stars on the feedback prompt as well as a free write text box to add personal remarks.

-  The character usage limit on the feedback prompt likely needs some debugging.
- The background of the feedback prompt will also need to be extended.

### Test Plan  
Before
Screenshot 2025-04-20 at 4 35 44 PM

After
Screenshot 2025-04-20 at 4 42 13 PM

 Check that the submit button does not illuminate until all three ratings are given
Check that firebase has new ratings

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/includes/FeedbackPrompt.tsx
+++ b/src/components/includes/FeedbackPrompt.tsx
@@ -10,12 +10,14 @@ import IconButton from '@material-ui/core/IconButton';
 import CloseIcon from '@material-ui/icons/Close';
 
 type Props = {
-    onClose: (rating?: number, feedback?: string) => void;
+    onClose: (rating1?: number, rating2?: number, rating3?: number, feedback?: string) => void;
     closeFeedbackPrompt: () => void;
 };
 
 const FeedbackPrompt = (props: Props) => {
-    const [rating, setRating] = useState<number | null>(0);
+    const [rating1, setRating1] = useState<number | null>(0);
+    const [rating2, setRating2] = useState<number | null>(0);
+    const [rating3, setRating3] = useState<number | null>(0);
     const [feedback, setFeedback] = useState<string>("");
     const containerRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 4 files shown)


src/components/includes/FeedbackPrompt.tsx
```diff
@@ -78,22 +107,41 @@ const FeedbackPrompt = (props: Props) => {
...
+                    onChange={(event) => {
+                        const input = event.target.value;
+                        if (input.length <= FEEDBACK_CHAR_LIMIT) {
+                            setFeedback(input);
+                        }
+                    }}
                 />
+                <Typography
+                variant="caption"
+                color={feedback.length >= FEEDBACK_CHAR_LIMIT ? "error" : "textSecondary"}
```
NIDHI2023, line 120: Make sure to run `npx eslint src --ext tsx --ext ts --fix` to fix the lint errors, or just manually adjust the spaces

NIDHI2023 (commented):
Good work! Make sure to merge in master and also update the fireData.d.ts file since you're changing the structure of how we store the feedback in the firebase. I'll do some more detailed testing later

src/components/types/fireData.d.ts
```diff
@@ -140,8 +140,8 @@ interface FireCourse {
     year: string;
     timeLimit?: number;
     timeWarning?: number;
-    isTimeLimit?: boolean;
-    feedbackList?: FeedbackRecord[]; // TODO: possibly change to non-null
+    isTimeLimit?: boolean; // TODO: possibly change to non-null
+    feedbackList?: FeedbackRecord[];
```
NIDHI2023, line 144: Since you're storing it in "users" now, I think you could remove this from here

src/components/includes/FeedbackPrompt.tsx
```diff
@@ -78,31 +113,48 @@ const FeedbackPrompt = (props: Props) => {
                     fullWidth
                     style={{ marginBottom: "4rem" }}
                     placeholder="Please describe your experience..."
-                    onChange={(event) => setFeedback(event.target.value)}
+                    /* Adds a character limit to the feedback response */
```
NIDHI2023, line 116: I think you use should the maxRows={4} or maxRows ={5} here so it becomes a scrollable text view instead of overflowing!

src/components/includes/FeedbackPrompt.tsx
```diff
@@ -78,31 +113,48 @@ const FeedbackPrompt = (props: Props) => {
...
+                    onChange={(event) => {
+                        const input = event.target.value;
+                        if (input.length <= FEEDBACK_CHAR_LIMIT) {
+                            setFeedback(input);
```
NIDHI2023, line 120: And I think like we talked about, make sure when it hits the 0 character limit implement the logic so it doesn't allow you to type more

src/components/includes/FeedbackPrompt.tsx
```diff
@@ -39,70 +44,119 @@ const FeedbackPrompt = (props: Props) => {
...
+                            setRating2(newValue);
+                        }}
+                        size="large"
+                        emptyIcon={<StarBorderIcon fontSize="inherit" />}
+                    />
+                    <br />
+                </Box>
+                <Typography variant="body2" style={{ fontStyle: "roboto", fontSize: "14px" }}>
+                    Overall Experience
+                </Typography>
```
anchen9, line 95: Hi Nikhill! The student feedback prompt component looks great! The flow went smoothly on my end and the submit rating only highlighted when every rating was filled! 

One suggestion here is I'm wondering whether it's possible to consolidate these rating blocks by using `.map()`! For example, you could have a list of the rating labels like `["Organization", "Efficiency", "Overall Experience"] `and map over these labels to render each rating as a small reusable component!

PR #957, opened 2025-05-04
TA Dashboard - Analytics Page

anchen9:
### Summary 

Implement the TA Dashboard Analytics Page following this [ticket](https://www.notion.so/TA-Dashboard-Frontend-Analytics-Tab-1b90ad723ce180d0bf79e3af085da5d4) and this figma [file](https://www.figma.com/design/gKqRwgffXSKNB5PYQlXHS0/Sia-FA-24-Working-File----Renee-SP25-Work-?node-id=1-124&p=f&m=dev).

- implement eric's components
- added separate scss files
- implement student reviews component
- implement metrics graphs

### Test Plan 

- test the filtering
- will adjust dummydata to test for different filtering and percentage bar displays
- will adjust dummydata to test for graph appearance with edge cases

### Notes 

- future considerations: 
- adjust/add hover state based on design feedback
- potentially adjust tooltip styling 
- might change based on backend implementation
- still need to refactor code & address some comments from prev pr oops ;( will be doing asap

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/components/includes/TAStudentTrends.tsx
```diff
@@ -0,0 +1,146 @@
...
+    // set tasks default to most recent assignment
+    const [sortFilter, setSortFilter] = useState("Query Volume");
+    const [filteredData, setFilteredData] = useState(dummyData);
+
+    useEffect(() => {
+        /* TODO: adjust filters after backend implementation to fully match figma. */
+        let result = [...dummyData];
+
+        // logic for the time dropdown filter
+        result = result.filter(item => {
```
EricWeng23, line 30: Good job Annie on these frontend implementations! I think it would be great to leverage the `useCallback` hook to wrap all of these filtering logics. The main reason behind this is because the filtering logic is quite complex and on every re-render it would have to recreate these filtering function instances even if the dependencies does not change. This is where `useCallback` can be used since it will return the same filtering functions as long as the dependencies does not change even across re-renders. An example of how I would optimize is would be:
```
const getFilteredData = useCallback(() => {
... your filtering logic ...
}, [... dependencies])
...
useEffect(() => {setFilteredData(getFilteredData());}, [getFilteredData])
```

EricWeng23 (dismissed):

src/components/includes/TAStudentTrends.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    // useStates to track filter selection
+    const [timeFilter, setTimeFilter] = useState("This week");
+    const [taskFilter, setTaskFilter] = useState(tasks[tasks.length - 2]);
+    // set tasks default to most recent assignment
+    const [sortFilter, setSortFilter] = useState("Query Volume");
+    const [filteredData, setFilteredData] = useState(dummyData);
+
+    useEffect(() => {
+        /* TODO: adjust filters after backend implementation to fully match figma. */
```
EricWeng23, line 26: Good job Annie on these frontend implementations! I think it would be great to leverage the `useCallback` hook to wrap all of these filtering logics. The main reason behind this is because the filtering logic is quite complex and on every re-render it would have to recreate these filtering function instances even if the dependencies does not change. This is where `useCallback` can be used since it will return the same filtering functions as long as the dependencies does not change even across re-renders. An example of how I would optimize is would be:
```
const getFilteredData = useCallback(() => {
... your filtering logic ...
}, [... dependencies])
...
useEffect(() => {setFilteredData(getFilteredData());}, [getFilteredData])
```

src/components/includes/StudentReviewPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+                <div className="header-text-container">
+                    <p className="header-text">Student Reviews </p>
+                    <p className="review-count">{`(${sortedReviews.length})`}</p>
+                </div>    
+                <FilterDropdown />
+            </div>
+            <div className="reviews">
+                {sortedReviews.map((review) => {
+                    return (
+                        <StudentReviewCard
```
NikhillA, line 73: Great job Annie! I love the look of the dashboard and especially like bar graph metrics at the top. The only thing I am wondering is that for the student reviews, it seems as though the reviews include decimal ratings like 3.5 and currently, the feedback prompt only allows students to give whole number ratings (1-5 stars). We can connect on this to see what would work best later.

NIDHI2023:
Amazing work as always! It looks just like the Figma designs, it's so impressive you can follow them so closely. For responsive layout, since you're using grids, I think this video will be helpful: https://www.youtube.com/watch?v=3T0gjtXRNC0

PR #958, opened 2025-05-05
Nikhill/llm

NikhillA:
### Summary <(In Progress PR) Student Feedback - integrating an LLM>

 Implemented character limit on feedback prompt so that users cannot abuse the free response section of the feedback prompt. Added another field (verified) to submitFeedback portion of Firebase to keep track of which reviews have been verified and which have not. Also estimated how many tokens would be passed through a typical feedback response to use in the LLM.

Over the summer, I will continue working on this PR by 1. Installing a Gemini library. 2. Running some prompts to get a deterministic output. 3. Test model locally to fine tune prompting and reviews.

 

### Test Plan 

 Run a test of a list of all bad reviews to see how many reviews the model is able to capture. Run again with neutral and good reviews. 

### Breaking Changes 

None

- [] I have updated the documentation accordingly.
- [] My PR adds a @ts-ignore


src/firebasefunctions/sessionQuestion.ts
```diff
@@ -242,36 +224,47 @@ export const getComments = (questionId: string, setComments: ((comments: FireCom
...
-            session,
-            questionId: removedQuestionId,
-            rating,
-            writtenFeedback: feedback,
+/* Adds three new ratings (organization, efficiency, overallExperience) 
+to the firebase under the users tab for each question asked. Also adds a verification field
+to determine whether or not a review was checked. */
+export const submitFeedback =
+    (removedQuestionId: string | undefined) =>
+        (rating1?: number, rating2?: number, rating3?: number, feedback?: string, verified?: boolean | undefined) => {
```
NIDHI2023, line 234: I'm a little confused how this is supposed to be called - it looks like a function inside a function? When you call submitFeedback in SessionView I think you call `submitFeedback(removedQuestionId)` but FeedbackPrompt is called with the other three parameters. Is there a way to make this cleaner, maybe just one function that takes four arguments?

src/components/includes/FeedbackPrompt.tsx
```diff
@@ -1,21 +1,26 @@
...
-    onClose: (rating?: number, feedback?: string) => void;
+    onClose: (rating1?: number, rating2?: number, rating3?: number, feedback?: string) => void;
     closeFeedbackPrompt: () => void;
 };
 
 const FeedbackPrompt = (props: Props) => {
-    const [rating, setRating] = useState<number | null>(0);
+    const [rating1, setRating1] = useState<number | null>(0);
+    const [rating2, setRating2] = useState<number | null>(0);
+    const [rating3, setRating3] = useState<number | null>(0);
```
laylaliuuu, line 23: Maybe make it a little bit clear what's really rating1, rating2, rating3

src/components/includes/FeedbackPrompt.tsx
```diff
@@ -26,10 +31,10 @@ const FeedbackPrompt = (props: Props) => {
             }
         };
 
-        document.addEventListener('mousedown', handleClickOutside);
+        document.addEventListener("mousedown", handleClickOutside);
 
         return () => {
-            document.removeEventListener('mousedown', handleClickOutside);
+            document.removeEventListener("mousedown", handleClickOutside);
```
laylaliuuu, line 37: Good debug catch

src/components/includes/FeedbackPrompt.tsx
```diff
@@ -39,70 +44,119 @@ const FeedbackPrompt = (props: Props) => {
...
                     minRows={4}
+                    maxRows={5}
                     fullWidth
                     style={{ marginBottom: "4rem" }}
                     placeholder="Please describe your experience..."
-                    onChange={(event) => setFeedback(event.target.value)}
+                    /* Adds a character limit to the feedback response */
+                    onChange={(event) => {
+                        const input = event.target.value;
+                        const length = input.length;
```
laylaliuuu, line 120: Good comment here, maybe add more comments like this would be helpful

src/components/types/fireData.d.ts
```diff
@@ -140,8 +140,8 @@ interface FireCourse {
     year: string;
     timeLimit?: number;
     timeWarning?: number;
-    isTimeLimit?: boolean;
-    feedbackList?: FeedbackRecord[]; // TODO: possibly change to non-null
+    isTimeLimit?: boolean; // TODO: possibly change to non-null
```
laylaliuuu, line 143: nice documentation.

laylaliuuu (commented):
very good overall. Nice clean code. I like the comments/documentation, maybe add more for major features would be very helpful. Maybe also consider making the variable name more specific, so it's easier to understand/

PR #938, opened 2025-03-24
layla/OH Time Graph Frontend

laylaliuuu:
### Summary 
New OH Time Graph Frontend, intend for students to have a sense of average wait time at different times of day, times. 

Design before:
image

Design after (goal):
image
but wait time section references to this:
image

Figma Design Link:
https://www.figma.com/design/WFhJLWyi7EbCP08jKrTXxc/Lucy-FA24-Working-File?node-id=426-2293&m=dev

- Screenshot of the current code design: 
image
image
image

### Test Plan 

- Hover states works on TA's profile pic
- Add other TAs, does information display in frontend (pic + ratio + arrow)
- Change the office hour title, location, it should update on frontend

### Notes 

Future To Do List: 

1. Replace Location section so the “Online” section with a link or a textbox if the student has to provide a link. Try to make it clearer whether the TA or Student has to provide the link
There are 2 different cases:
- Online: 1) Student view: if the TA didn't provide a zoom link, have a textfield under the time/date (left top box), and prompt the student to enter a zoom link (can reference back to current website for zoom). TA view: to provide a link, or use student's provided link. 2)Student view: if the TA provided a zoom link, just have a zoom icon and says "Join", if not provided, then do a textfield and enter their own zoom link. TA view: Provide a zoom link (with textfield), or have "zoom icon and says "Join"
- Hybrid: the logic same as Online, but the current QMI is not super clear. 
- Make sure the OH Info makes sense for Virtual and Hybrid modes as well (ex: if the session is virtual, you shouldn’t have a location like Gates G21)

2. Mobile and desktop views (responsive design) 

3. (DONE) Waittime Graph (use Eric's backend information and replace the dummy data), Eric made a PR can reference to that (https://github.com/cornell-dti/office-hours/pull/953). And make sure to fix the wait time section design to match with the picture above. Add this chalkboard and hourglass icon according to Renee's designs, so they should be in two separate lines. 

4. Replace Dummy data for wait time graph. 

5. (DONE) Add hover (TA's full name) to their profile picture. 

### Breaking Changes 

None

- [ ] I have updated the documentation accordingly.
- [ ] My PR adds a @ts-ignore

PR #967, opened 2025-09-22
Adding Zoom Details to New UI

anchen9:
### Summary 

Implemented part 1 of this [ticket](https://www.notion.so/OH-Time-Graph-Frontend-Mobile-view-Zoom-link-fixes-2670ad723ce18015800bc0bb82d35014). This PR essentially adds the Zoom UI (svg, link, join button, etc.) back to the new UI from the old UI for the SessionInformationHeader file. Continuation of Layla's PR #938 

Some cases covered:
- Case 1: Course Zoom Link, Virtual
Screenshot 2025-09-21 at 9 37 19 PM

- Case 2: TA Zoom Link, Virtual
Screenshot 2025-09-21 at 9 38 54 PM

Other cases covered: hybrid (student provide zoom link), student view for all of the above cases, etc.

### Test Plan 

- Test the cases listed in the notion ticket and what the view looks like from Prof, Student, TA side.

### Notes 

- some of the padding is a bit weird due to the MUI grid, could consider playing around with it further to adjust it.

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


src/components/includes/SessionInformationHeader.tsx
```diff
@@ -96,482 +100,539 @@ const SessionInformationHeader = ({
...
-                                                                value={zoomLink}
-                                                                onChange={e => setZoomLink(e.target.value)}
-                                                            />
-                                                            <div className="CloseZoom">
-                                                                <img
-                                                                    onClick={closeZoomLink}
-                                                                    src={closeZoom}
-                                                                    alt="close zoom"
+                                                    {zoomLinkDisplay === 'show' && (
+                                                        <>
```
NIDHI2023, line 290: Maybe this Grid could also be wrapped with justifyContent="flex-start" like the zoom button and then add an offset? So basically the zoom link stays a fixed margin away from the zoom link button. Currently at bigger widths, the link goes to the center while the zoom button stays on the left, so I think we should fix that spacing a bit

PR #969, opened 2025-09-22
Prompt to Gemini LLM to analyze question documents into tags (in Progress)

EricWeng23:
…o tags

### Summary 

This PR is the first step to initiate the waittime model to calculate individual question waittime in a session. In the previous attempt in PR https://github.com/cornell-dti/office-hours/pull/941, I had dummy data that reflected real question document fields from firebase. However, just relying on those fields does not generalize the model to give accurate predictions across different session where different classes can have longer waittime or more intense questions. Thus, I decided to move away from the Bag of Words approach and instead try to prompt a LLM like Gemini (in this case) to categorize the questions into tags like `complexity`, `type`, `topic`, `estimatedTimeRange` before we train a model. This is just preprocessing the Firebase question document into features that can be relevant. 

Checkout https://ai.google.dev/gemini-api/docs/quickstart#javascript

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/scripts/busyness/question_parser.ts
+++ b/src/scripts/busyness/question_parser.ts
@@ -1,0 +1,122 @@
+import { GoogleGenAI } from "@google/genai";
+import { firestore } from "../../firebase";
+import { doc, getDoc } from "firebase/firestore";
+
+/**
+ * Firebase Question Document Analyzer using Google Gemini API
+ * Analyze an incoming question document request adn categories the question by complexity, type, topic, and estimate time interval
+ */
+
+interface QuestionData {
+    content: string;
+    position: string | number;
+    status: string;
+}
+
+// Import FireQuestion type
+declare interface FireQuestion {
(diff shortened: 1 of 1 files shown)


src/scripts/busyness/question_parser.ts
```diff
@@ -0,0 +1,122 @@
+import { GoogleGenAI } from "@google/genai";
```
anchen9, line 1: Hi Eric! Great work on implementing this and also on demoing this at work sesh! Aside from a couple lint fixes, the logic behind your code makes sense! I was also just curious as to whether the "@google/genai" import needed to also be added to the package.json?

NikhillA: I had this same error on my end and turns out we need to download gemini without changing the office hour dependencies. Once that happens, this error gets resolved.

PR #975, opened 2025-10-05
replace dummy data

laylaliuuu:
### Summary 

Replaced dummy data, and create variables. 

### Test Plan 
Configure firebase and test some courses

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


scripts/setup-historical-data.ts
```diff
@@ -0,0 +1,129 @@
...
+        console.log(`✅ Updated ${questionsSnapshot.docs.length} questions for session ${sessionId}`);
+    }
+    
+    console.log(`📊 Total questions updated: ${totalUpdated}`);
+};
+
+/**
+ * Calculate historical wait times for all courses
+ */
+const calculateHistoricalDataForAllCourses = async () => {
```
NikhillA, line 84: Great job frequently using console.log for different actions throughout this file! It makes understanding your work a lot easier.

NikhillA:
Great job on this PR Layla! Nice job testing with some courses from firebase and changing out the dummy data. Your code is really easy to read and understand functionally!

src/components/includes/SessionInformationHeader.tsx
```diff
@@ -145,65 +140,53 @@ const SessionInformationHeader = ({
...
-        } else {
-            message = `Please wait for your turn to join the Zoom call (estimated wait time: ${avgWaitTime})`;
+            const date = new Date(selectedDateEpoch);
+            const exists = await hasSessionsOnDate(course.courseId, date);
+            if (!ignore) setHasSessionsSelectedDay(exists);
         }
-        setShowErrorMessage(message);
-    };
+        run();
+        return () => {
```
NIDHI2023: I'm a little unsure if this return does much? Doesn't ignore always get set to false at the start of the useEffect?

src/components/types/fireData.d.ts
```diff
@@ -146,6 +146,11 @@ interface FireCourse {
     timeWarning?: number;
     isTimeLimit?: boolean;
     feedbackList?: FeedbackRecord[]; // TODO: possibly change to non-null
+    historicalWaitTimes?: {
```
NIDHI2023: It looks like in the firebase this is called waitTimeMap
image

Maybe change the name here to be consistent with that?

NIDHI2023 (commented):
Awesome work Layla, especially with figuring out the edge cases of the time graph and choosing a structure to store the data in! J

ust a comment, I'm not exactly sure where you're calling the functions to generate the data in the firebase, but I would suggest creating the null values for the waitTimeMap field only for courses in the current sem/ when a new course is created so that when we deploy, we aren't doing a ton of reads for courses we don't use anymore

PR #976, opened 2025-10-05
Add courseId field to questions documents

NIDHI2023:
- this fixes tags bar chart bug, where getQuestionsQuery in firehooks.ts wasn't valid because we didn't have this field. note that this query will only return documents that have courseId going forward
- also fixed percentage text for long decimals
- should help queries for Layla and Annie's work to get question statistics

image

### Notes 

- could cause a lot of reads whenever someone opens the analytics pages, but I don't think we can avoid that since we have to iterate through all the questions in a course to get this info for specific tags

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore

### DIFF
--- a/src/components/pages/ProfessorDashboardView.tsx
+++ b/src/components/pages/ProfessorDashboardView.tsx
@@ -130,7 +130,10 @@ const ProfessorDashboardView = ({ match: { params: { courseId } } }: RouteCompon
                                     <p className="totalQuestions"> {currentCategory.totalQuestions} </p>
                                     <p className="totalQuestionsLabel"> questions total </p>
                                     <hr />
-                                    <p className="percentResolved"> {Number.isInteger(currentCategory.percentResolved) ? currentCategory.percentResolved: currentCategory.percentResolved.toFixed(2)}% </p>
+                                    <p className="percentResolved"> {Number.isInteger(currentCategory.percentResolved)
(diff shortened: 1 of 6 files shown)


src/components/pages/ProfessorDashboardView.tsx
```diff
@@ -130,7 +130,10 @@ const ProfessorDashboardView = ({ match: { params: { courseId } } }: RouteCompon
                                     <p className="totalQuestions"> {currentCategory.totalQuestions} </p>
                                     <p className="totalQuestionsLabel"> questions total </p>
                                     <hr />
-                                    <p className="percentResolved"> {currentCategory.percentResolved}% </p>
+                                    <p className="percentResolved"> {Number.isInteger(currentCategory.percentResolved)
+                                        ? currentCategory.percentResolved:
```
laylaliuuu: Nice logic here

src/firebasefunctions/sessionQuestion.ts
```diff
@@ -44,6 +45,7 @@ export const addQuestion = (
             ...upvotedUsers,
             ...addVirtual,
             answererId: '',
+            courseId:course.courseId,
```
laylaliuuu: good getting courseID here

laylaliuuu (commented):
Code is very clean and clear. Maybe adding some comments would be helpful

PR #1008, opened 2026-01-07
Finalize TA Dashboard

NIDHI2023:
### Summary 
This PR makes some revisions to PR #991 to finalize the student review feedback feature for both frontend and backend. It also finishes up PR #1001 by creating a weekly scheduled Firebase Cloud function to generate student trends, instead of running it manually. It also fixes the ordering of the "first mentioned" filter.
Before:
image

Video of student side and bad feedback not getting stored:

https://github.com/user-attachments/assets/84c4ec8d-1501-47b0-8d5b-53261bf9e5cc

(the other bad feedbacks in the collection were tests)

Image of TA side:
image
image

Image of successful clustering and filtering generated from the scheduled function:
image

### Test Plan 

- Tested student entering good and bad feedback (function logs notify when feedback is rejected)
- Tested that TA sees only good feedback in dashboard, accurate ratings and feedback content
- TA Dashboard updates on refresh when new feedback added
- Tested that titles and clustering generally make sense
- Tested that scheduled function successfully ran

### Notes 

- **To get Genkit working, I had to enable the Vertex AI API in the google cloud project & add IAM permission to the default compute service account as a Vertex AI User role.**  The Gemini API key was not working (weird organization restrictions). 

- Don’t need to wrap verification flow with onCallGenkit (a v2 cloud function) since everything is server side, not https calls. I also found that with mixing v1 and v2 in the same index.ts file, you cannot use function.config, so we would need to delete the twilio code and use parameterized configuration (https://github.com/firebase/firebase-tools/issues/6465). You will end up with a separate env file when deploying, I ran `firebase functions:config:export`

- I refactored the architecture to use a feedback subcollection for each user instead of onScheduled (another v2 cloud function, also was hard to test), which allowed easier cloud function triggers to run the verification and easier logging. **When a new feedback record is added to feedback, the function calls the API to immediately verify it. If it determines it is inappropriate, the document is deleted (which saves storage).** The logic is also easier to follow instead of having verification fields in a feedback list and iterating through that each time.

- Also changed ta metrics to use local day instead of UTC day since that made more sense to me (otherwise I was getting confused why my questions for tuesday were showing up as wednesday)

- Instead of running Annie's clustering work on all the questions every week, I decided to save older work and only run it on new questions for that week. This should still cover all questions asked since the function should run every week. While it requires some extra checks to merge new and old questions under the same titles, I believe it's cheaper than running costly ML functions on a larger growing set of questions and not using any of the old work.

- I set the memory to be 2GB just in case there would be a lot of questions, but in production if it's able to run with less memory, that would also be cheaper

### Breaking Changes 

- Deployed new cloud function, enabled APIs
- Added security rules for feedback subcollection
- Added genkit and @genkit-ai/google-genai packages in functions
- Added new index for questions collection on courseId and timeEntered

- [ ] My PR adds a @ts-ignore


nadia-choop (approved):
Thanks Nidhi, this looks great! Really appreciated the approach to using previous analytics as well to reduce cost going forward.

PR #1050, opened 2026-09-03
add genkit to package.json

plumshum:
### Summary 

the codebase also imports directly from `genkit`, not just `google-ai/genkit`, so i also added the dependency in the package.json 

### Breaking Changes 

None

- [ ] My PR adds a @ts-ignore


shah-esha (approved):
lgtm

